PR #81, opened 2015-07-22
Improve translation loading

ragulka:
Currently a plugin that implements the framework also includes all framework strings in it's POT file, which means there's duplicated strings in every plugin. We should try to separate the framework strings into their own POT file so it can be translated once and loaded independently.

It's also possible that this would allow us to use strings for text domains again instead of making use of class constants and class members.

specific improvements:

* [payment token names](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/class-sv-wc-payment-token.php#L243-L250) - no access to the plugin text domain without overriding the class, but a `wc-plugin-framework` text domain would make them translatable

for reference:
* https://github.com/woothemes/woocommerce/issues/2051
* https://developer.wordpress.com/2015/04/23/wordpress-developers-test-your-i18n-internationalization-knowledge/
* http://ottopress.com/2012/internationalization-youre-probably-doing-it-wrong/


ragulka:
Closing in favour of #90

PR #90, opened 2015-07-23
Refactor translation loading

ragulka:
Originally from #81. Opening a new PR against the 4.0-beta branch for easier comparison of changes.

Original issue description by @maxrice:

Currently a plugin that implements the framework also includes all framework strings in it's POT file, which means there's duplicated strings in every plugin. We should try to separate the framework strings into their own POT file so it can be translated once and loaded independently.

It's also possible that this would allow us to use strings for text domains again instead of making use of class constants and class members.

specific improvements:

payment token names - no access to the plugin text domain without overriding the class, but a wc-plugin-framework text domain would make them translatable
for reference:

woothemes/woocommerce#2051
https://developer.wordpress.com/2015/04/23/wordpress-developers-test-your-i18n-internationalization-knowledge/
http://ottopress.com/2012/internationalization-youre-probably-doing-it-wrong/

~~**This PR breaks backwards-compatibility**~~ Well, no, it doesn't. Unless we change the plugin constructor signatures.

**Ideas**

- [ ] Integrate translation workflow with Transifex

**Todo in wc-plugins**

- [ ] Plugins need to be updated to use a hardcoded text-domain (string literal) instead of a constant
- [ ] Plugins need to be updated to use correct constructor signatures, see 2b99b7d6c0cb962f4c9abbe76bf3185986e6f1ac
- [ ] Plugin translations should be refactored based on the new [translation standards](https://github.com/skyverge/wc-plugins/wiki/Translations)
- [ ] Grunt makepot task needs to be updated to exclude framework files
- [ ] Grunt build/deploy task needs to be updated to exclude framework Gruntfile.js, package.json, node_modules, .gitignore, etc.
- [ ] Perhaps introduce a `update_translations` task for plugins, similar to this: https://github.com/skyverge/wc-plugin-framework/blob/translations/Gruntfile.js#L51-L71


woocommerce/class-sv-wc-plugin.php
```diff
@@ -198,7 +201,23 @@ public function __clone() {
...
+
+	/**
+	 * Load plugin & framework text domains
+	 *
+	 * @since 4.0.0-beta
+	 */
+	public function load_translations() {
+
+		// Load framework text domain
+		load_plugin_textdomain( 'sv-wc-plugin-framework', false, dirname( plugin_basename( $this->get_file() ) ) . '/lib/skyverge/woocommerce/i18n/languages' );
```
maxrice: this should probably use `$this->get_framework_path()` since we want to load translations from the currently loaded framework path

woocommerce/class-sv-wc-plugin.php
```diff
@@ -564,7 +586,7 @@ public function get_admin_notice_handler() {
 			return $this->admin_notice_handler;
 		}
 
-		return $this->admin_notice_handler = new SV_WC_Admin_Notice_Handler( $this, $this->text_domain );
+		return $this->admin_notice_handler = new SV_WC_Admin_Notice_Handler( $this, 'sv-wc-plugin-framework' );
```
maxrice: does this still need a text domain passed in?

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-user-edit-handler.php
```diff
@@ -56,6 +56,8 @@ class SV_WC_Payment_Gateway_Admin_User_Edit_Handler {
 	/**
 	 * Initialize and setup the Admin User Edit Handler
 	 *
+	 * TODO: $text_domain is deprecated, remove argument on next breaking release
+	 *
 	 * @since 3.0.0
 	 */
 	public function __construct( $plugin, $text_domain ) {
```
maxrice: think the text domain param for this can be removed :)

maxrice:
@ragulka this looks great overall, a few questions:

* are there any issues with loading 2 text domains in a given plugin? I seem to recall WooCommerce fighting with this, but maybe it's no big deal.

* any reason for the `sv-wc-plugin-framework` text domain over `wc-plugin-framework`?

* Aside from the constructor param change, I don't see any other breaking changes here. `load_translations()` in plugins would need to be updated to call parent before loading their own though.

Once we've got these figured out, I'll merge this into #92 so we can keep iterating it on for the next release. thanks!

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-user-edit-handler.php
ragulka: Right, that's what I did originally in 2b99b7d6c0cb962f4c9abbe76bf3185986e6f1ac, but reverted it in ae1b0412fb4260ae2cbc7dd2cc9b2f8c22d51d50 so that this would not be a breaking change. We discussed that in Slack: https://skyverge.slack.com/archives/woocommerce/p1437688625001115

maxrice: apologies, after thinking about what we're doing for the 4.1 milestone, we can break backwards compat for gateways but not for all plugins, so in this specific intstance it's fine to pull the `$text_domain` param out :)

woocommerce/class-sv-wc-plugin.php
ragulka: Not really :)

ragulka:
> are there any issues with loading 2 text domains in a given plugin? I seem to recall WooCommerce fighting with this, but maybe it's no big deal.

AFAIK, no. Even Codestyling seems to support it: https://wordpress.org/support/topic/multiple-textdomains-in-a-single-plugin?replies=6

> any reason for the sv-wc-plugin-framework text domain over wc-plugin-framework?

convention? :) all the classes are prefixed with SV... we can remove that if you want?

> Aside from the constructor param change, I don't see any other breaking changes here. load_translations() in plugins would need to be updated to call parent before loading their own though.

Actually, they don't. There are 2 methods now: `load_translations()` (plural) and `load_translation()` (singular). I only introduced the plural method - it loads the framework translation and then calls `load_translation()`. No need to call any parent methods :)

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-user-edit-handler.php
ragulka: Done :)

maxrice:
@ragulka oops, didn't mean to have this close. I've started merging the commits on this into the 4.1 branch (see #92) and will have that finished tomorrow. Once that's done, we decided on "WooCommerce Plugin Framework" for the long title and `woocommerce-plugin-framework` for the text domain.

Can the transifex config live outside of that .tx folder or must it go there?

maxrice:
@ragulka merged the commits and fixed up all the conflicts except for the gruntfile. either you or @tamarazuk can do that since I'm not sure what the correct structure is. Please continue dev over in #92 :) thanks!

tamarazuk:
@maxrice @ragulka merge conflicts were fixed in SHA 129576b1d3b98dfad4cd8e69b14a82a98dd4f84c. @ragulka we adopted a non-standard method for organizing our grunt related code - tasks and configs are in separate files - which I applied here and am hoping is easy to follow, but let me know if you have questions. However, I wasn't able to run the `update_translations` task, is there a dependency I'm missing?

ragulka:
> Can the transifex config live outside of that .tx folder or must it go there?

@maxrice I don't think it can be moved: http://docs.transifex.com/client/config/#txconfig

> However, I wasn't able to run the update_translations task, is there a dependency I'm missing?

@tamarazuk did you see any particular errors? you need to have the `gettext` package installed, which means you need to have `xgettext`, `msgmerge`, `msgfmt` commands available on your system.

PR #92, opened 2015-07-28
4.2 Release

maxrice:
# General Todos
- [x] Filter on the default checked state for the "save to account" checkbox [here](https://github.com/skyverge/wc-plugin-framework/blob/wc-plugin-framework-4.0.1/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php#L682)
- [x] Inline docs for filter/actions
- [X] Update usages of `SV_WC_Payment_Gateway_Payment_Token::get_token()` with `SV_WC_Payment_Gateway_Payment_Token::get_id()
- [x] Document that the the backwards compatible flag in the FW boostrap means other frameworked plugins that are at the version specified can use this new version without issues
- [x] See [4.2 milestone](https://github.com/skyverge/wc-plugin-framework/issues?q=is%3Aopen+is%3Aissue+milestone%3A4.1) for additional issues.
- [x] Change the bootstrap message when another plugin requires an update to indicate that it is currently out-of-date, e.g. `The following plugin is inactive because it is out-of-date and requires a...`

# Translation Todos
- [x] Integrate translation workflow with Transifex
- [x] Plugins need to be updated to use a hardcoded text-domain (string literal) instead of a constant
- [x] Plugins need to be updated to use correct constructor signatures, see 2b99b7d6c0cb962f4c9abbe76bf3185986e6f1ac
- [x] Plugin translations should be refactored based on the new [translation standards](https://github.com/skyverge/wc-plugins/wiki/Translations)
- [x] Grunt makepot task needs to be updated to exclude framework files
- [x] Grunt build/deploy task needs to be updated to exclude framework Gruntfile.js, package.json, node_modules, .gitignore, etc.
- [x] Perhaps introduce a `update_translations` task for plugins, similar to this: https://github.com/skyverge/wc-plugin-framework/blob/translations/Gruntfile.js#L51-L71

# Meta
- [x] set license and improve readme (remove "known issues" section as it no longer applies)
- [x] consider adding sample implementations (general plugin, with an API class implementation, and a sample payment gateway implementation)
- [x] Draft blog post

# Done
- [X] Payment form feature accepts a card expiration date like `01-2016` in addition to the correct format like `01-16`. We should convert the longer form internally to `01-16` so gateway implementations can rely on a consistent format. SHA 041c8aa87fe38f161fb1c27bc04190ec67e37a12
- [X] Change [this](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php#L642) to 2 digit year SHA 041c8aa87fe38f161fb1c27bc04190ec67e37a12
- [x] Run a code inspection with PhpStorm


maxrice:
@justinstern see https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/api/interface-sv-wc-payment-gateway-api-payment-notification-response.php#L71 -- do you still think this should be moved? It doesn't seem like a majority of payment gateway API calls would return the payments type

justinstern:
> see https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/api/interface-sv-wc-payment-gateway-api-payment-notification-response.php#L71 -- do you still think this should be moved? It doesn't seem like a majority of payment gateway API calls would return the payments type

@maxrice hmm, well I don't think it matters whether the payment gateway API call itself returns the payment type, that's something that the response object should always be able to know. 

So to me the question is whether it's potentially useful having that available via a convenience method. It seemed like a reasonably generic enough thing to have defined on the parent class rather than just the hosted gateway response, but perhaps it's just not useful in a direct gateway, idk, I'm fine either way on it

ragulka:
@maxrice the translations part is now ready to be tested.

There are 2 grunt tasks worth mentioning:

`grunt update_translations` - will:
1. extract strings from source files and generate the POT file
2. upload it to Transifex
3. download updated translations from Transifex
4. post-process PO file headers to include framework version number (something that Transifex does not do)
5. Compile PO files to MO files

`grunt build` - will:
1. Compile & minify coffee/JS
2. Compile SCSS
3. Update translations (see above)

As it stands, the grunt tasks will pull the framework title and version from package.json. This means that before releasing, make sure to update the version number in package.json as well as other files, where necessary.

@tamarazuk - when you have time, it would be great if we had version bump, deploy etc tasks for the framework as well.

tamarazuk:
> when you have time, it would be great if we had version bump, deploy etc tasks for the framework as well.

@ragulka I'll work on this today or tomorrow :)

PR #94, opened 2015-07-29
get_default_title return consistency

tivnet:
Function is described as `string`, so it must not have a `void` return.


maxrice:
merged into 4.1 with SHA 26d59396d41a148e48bf50d9e37f4eac81fff428

PR #95, opened 2015-07-29
validate_fields return consistency

tivnet:
- `$this->$method_name( $is_valid );` looks like missing return?
- default return is missing, too. I guess, it's `true`?


justinstern:
@tivnet good catch here. Just change that final `return true;` to `return $is_valid;` so that we're returning whatever the parent method returned, by default

tivnet:
@justinstern ops. Of course, `return $is_valid;`. Thank you!

justinstern:
:+1:

maxrice:
merged into 4.1 with SHA 0c1db330f3b767aedb6632e333676e9325d2b0da

PR #97, opened 2015-07-29
get_default_description return consistency

tivnet:
Again, the method should not return `void`. Need to QA for (unlikely) side effects.


maxrice:
merged into 4.1 with SHA 20952a938c79b1080d891fa68d92dd7d54e77252, thanks

Issue #96, opened 2015-07-29
SV_WC_Payment_Gateway_Direct::process_payment


tivnet:
I am not sure how to PR this. Too many "unknowns".

- the parent method returns `array()`, so this method should return array, too
- `$result = apply_filters( 'wc_payment_gateway_'` is compared to `true` (?) and then there is return without checking what is returned. Should still be an array (?)
- nothing returned in `catch` and nothing returned by default at the end.

maxrice:
merged in SHA aa04bc81acadc5ac8c0f83aca19f475504550e5c

PR #98, opened 2015-07-31
[#96] Cleanup payment gateway process_payment() method

justinstern:
Match return value to method signature

see #96


maxrice:
merged in SHA aa04bc81acadc5ac8c0f83aca19f475504550e5c thanks

PR #100, opened 2015-08-05
Support per-order currencies

ragulka:
This change paves way for multi-currency support. At the current level, it simply allows gateways to:
* change the supported currencies runtime (use-case: country-specific currencies - if a gateway supports multiple countries and each country supports a different currency, then the gateway must be able to change the supported currencies runtime)
* use the order currency instead of configured WC currency, where available. If the order currency is different from the configured WC currency, this change will still allow paying with a gateway that supports the order currency. Without it, the gateway would not be available, even though the order is in a supported currency.

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -242,7 +242,7 @@ abstract class SV_WC_Payment_Gateway extends WC_Payment_Gateway {
 	private $available_card_types;
 
 	/** @var array optional array of currency codes this gateway is allowed for */
-	private $currencies;
+	protected $currencies;
 
 	/** @var string configuration option: the transaction environment, one of $this->environments keys */
 	private $environment;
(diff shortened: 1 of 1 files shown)


ragulka:
@justinstern @maxrice whaddayathink?

maxrice:
Looks fine to me, though this is only applicable for pay page gateways, no? What about an optional `$order` parameter for the `get_payment_currency()` method for direct gateways?

maxrice:
@ragulka merged, feel free to make additional commits on this directly to the 4.1 branch

ragulka:
> this is only applicable for pay page gateways, no?

Hmm, I'm not sure - I certainly did not design it like that. Are direct gateways not using the same method? Or is the global `$order` not available for them?

maxrice:
I might be misunderstanding how this would be used -- generally a direct gateway would process payment on the checkout page, not the pay page (though it certainly works there too)

ragulka:
Right, so the idea behind this commit is that if you have an order, which is pending payment under My Account. You go to My Account -> My Orders -> Select a pending order -> Pay. That's actually the only place where the `order-pay` query arg is available, and where the order currency can differ from the WC currency, AFAIK. In checkout, you are always using the `get_woocommerce_currency()`.

It should not matter if it's a hosted or direct gateway, since this should only affect whether the gateway is listed as an available option for payment or not.

maxrice:
>Right, so the idea behind this commit is that if you have an order, which is pending payment under My Account. You go to My Account -> My Orders -> Select a pending order ->

Ah right, so this is designed for pay page gateways, or paying a previously pending/failed order. That makes sense.

>In checkout, you are always using the get_woocommerce_currency().

With stock WC, yes. However, there are a few currency switchers that set the order currency at checkout. Handling that with `$order->get_order_currency()` works fine though.

ragulka:
> However, there are a few currency switchers that set the order currency at checkout. Handling that with $order->get_order_currency() works fine though.

Right, so this should be fine, then, no? Is there a plugin we could test this with, just to make sure?

maxrice:
WPML has one and there's also the [Aelia Currency Switcher](https://aelia.co/shop/currency-switcher-woocommerce/). I'll ask Diego for a copy.

PR #149, opened 2016-05-27
Introduce shared exporter classes

ragulka:
WIP, do not merge yet. 

Related to https://github.com/skyverge/wc-plugins/issues/647

### DIFF
--- a/woocommerce/class-sv-wc-plugin.php
+++ b/woocommerce/class-sv-wc-plugin.php
@@ -69,6 +69,9 @@ abstract class SV_WC_Plugin {
 	/** @var array string names of required PHP functions */
 	private $function_dependencies = array();
 
+	/** @var string the plugin text domain */
+	private $text_domain;
+
 	/** @var SV_WC_Admin_Notice_Handler the admin notice handler class */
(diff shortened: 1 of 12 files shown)


ragulka:
@maxrice @ChaseWiseman would love to have your initial feedback on this. I skipped abstracting anything admin/UI related at this point as I wanted to get the basic functional building blocks in place first, implement the column picker & background exports first and then see, if and what we can/want to abstract to the framework.

Basically, at the moment, I'd like to get a high-level "are we heading in the right direction?" with this. Basic code style, folder structure, naming, etc. Does that look OK?

Also, I've moved some stuff around to make the classes more modular. One thing I did was to remove any `get_option()` calls from the `Export_Method` classes, so that they would not expect having certain options in the db. This would make them more portable, and could be used in plugins such as Memberships to provide basic exporting functionality.

The `Cron` and `Export_Handler` classes are more dependent on loading data from options table and the plugin loading them, but if we wanted, we could abstract those as well, so that the options are passed in via the constructor or method signatures. I'm not yet convinced that this would bring any benefits though as it would mean that all the options are loaded upfront, not lazily, when needed.

PR #172, opened 2016-10-11
Object of class Requests_Utility_CaseInsensitiveDictionary could not be converted to string

sudiptochoudhury:
Fix issue related to PHP Catchable fatal error: Object of class Requests_Utility_CaseInsensitiveDictionary could not be converted to string in .../wp-content/plugins/woocommerce-gateway-authorize-net-cim/lib/skyverge/woocommerce/class-sv-wc-plugin.php on line 438

### DIFF
--- a/woocommerce/class-sv-wc-plugin.php
+++ b/woocommerce/class-sv-wc-plugin.php
@@ -469,7 +469,9 @@ abstract class SV_WC_Plugin {
 		$messages = array();
 
 		$messages[] = isset( $data['uri'] ) && $data['uri'] ? 'Request' : 'Response';
-
+		if (is_object( $value ) && 'stdClass' != get_class( $value )) {
+                	$value = json_decode($value);
+            	}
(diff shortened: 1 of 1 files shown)


ChaseWiseman:
thanks for the PR @sudiptochoudhury! 😄 

We addressed this in v4.4.1 of the framework, albeit using a different approach: https://github.com/skyverge/wc-plugin-framework/commit/109b030d02cf45c1d1ff363e57804417a61406e6

I will however take a look and see if there are ways we can improve log data handling in the future to avoid potential errors.

It looks like Authorize.Net CIM was triggering the error you saw. It's had this updated version of the framework since version 2.3.1

sudiptochoudhury:
Hi,

Thanks.

The fix applied on 4.4.1 seems fine.

I was using an older version of Authorize.net CIM plugin and had to add a
quick fix for a client.

Thanks again. Seems the pull request may be closed.






*Sudipto Choudhury*+91 98311 92043

PR #191, opened 2017-02-03
Delete token meta from admin editor regardless of API response

ChaseWiseman:
Previously, invalid tokens could not be removed using the admin token editor.

### DIFF
--- a/woocommerce/changelog.txt
+++ b/woocommerce/changelog.txt
@@ -1,5 +1,8 @@
 *** SkyVerge WooCommerce Plugin Framework Changelog ***
 
+2017.02.03 - version 4.5.3-dev
+ * Fix - Always delete a token from user meta when removed using the admin token editor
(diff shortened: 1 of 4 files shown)


maxrice:
Instead of a framework change here, I think we want to handle this on a per-gateway basis. Consider the following example using CIM:

- admin attempts to the delete the token from the editor
- temporary error with CIM causes the API request to fail
- the token appears to have been deleted
- when the tokens are refreshed from the CIM API again, the token is persisted

From the framework side, I think it'd be valuable to alert the admin if an error occurred during the deletion attempt (we're already returning `{ success: false }` JSON so that's sufficient to render an alert box). Then, for gateways that implement this they should handle any API response which indicates the token is missing/not found and we should delete it locally.

ChaseWiseman:
closing in favor of #200

PR #197, opened 2017-03-16
Add order date compatibility methods

ChaseWiseman:
Here we have it! Date compatibility methods for handling WooCommerce 3.0+ order dates.

### New methods:
- `SV_WC_Order_Compatibility::get_date_created( $order, $context = 'edit' );`
- `SV_WC_Order_Compatibility::get_date_modified( $order, $context = 'edit' );`
- `SV_WC_Order_Compatibility::get_date_paid( $order, $context = 'edit' );`
- `SV_WC_Order_Compatibility::get_date_completed( $order, $context = 'edit' );`

## Expected behavior

In all support WC versions, these `SV_WC_Order_Compatibility::get_date_*()` methods will return a `DateTime` object that you can use to manipulate the date format/output. For WC 3.0+, this object will be `WC_DateTime`. For WC 2.5.5 & 2.6.x, this adds the `SV_WC_DateTime` which mimics the latter so we can stick with a single implementation without version checks in the plugin.

## Implementation

As with our other 3.0 compatibility methods, these default to the `edit` context to avoid filters since previous uses of `$order->order_date`, etc... were similarly unfiltered. Be sure to look at the context of where you're using these. If it's decided that filtering is beneficial, you can set the `$context` param to `view`.

Here are a few ways to handle order dates using these methods.

```php
$order = wc_get_order( 123 );

// ISO 8601 date string
echo SV_WC_Order_Compatibility::get_date_created( $order );

// UTC timestamp
echo SV_WC_Order_Compatibility::get_date_created( $order )->getTimestamp();

// Timestamp offset to the shop timezone
echo SV_WC_Order_Compatibility::get_date_created( $order )->getOffsetTimestamp();

// Formatted date string, offset to the shop timezone
echo SV_WC_Order_Compatibility::get_date_created( $order )->date( 'Y-m-d' );

// Localized date string, offset to the shop timezone
echo SV_WC_Order_Compatibility::get_date_created( $order )->date_i18n( 'Y-m-d' );

// Any other format - http://php.net/manual/en/function.date.php
echo SV_WC_Order_Compatibility::get_date_created( $order )->format( 'c' );

// Order not completed? null - non-existent date
echo SV_WC_Order_Compatibility::get_date_completed( $order );
```

Also added is a helper method that makes it quicker to get a date string in the default WC format:
`SV_WC_Plugin_Compatibility::wc_format_datetime( $date );`

As we implement this in plugins we'll need to make sure and test across WC versions and not just in WC 3.0-RC.

### Question

Is the naming of the `SV_WC_DateTime` class too wonky? Shouldn't need to check for `WC_DateTime` specifically in plugins so I went with the SV namespace initially just in case, though it makes the docblocks a little funky since both `SV_WC_DateTime` & `WC_DateTime` could be accepted/returned depending on the WC version.

Would love some 👀 here on the code and QA to make sure everything's covered as it should be.

### DIFF
--- a/woocommerce/compatibility/class-sv-wc-order-compatibility.php
+++ b/woocommerce/compatibility/class-sv-wc-order-compatibility.php
@@ -51,6 +51,116 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
 
 
 	/**
+	 * Gets an order's created date.
+	 *
+	 * @since 4.6.0-dev
+	 *
+	 * @param \WC_Order $order order object
+	 * @param string $context if 'view' then the value will be filtered
+	 *
+	 * @return \WC_DateTime|null
+	 */
+	public static function get_date_created( WC_Order $order, $context = 'edit' ) {
+
+		return self::get_date_prop( $order, 'created', $context );
+	}
+
+
+	/**
+	 * Gets an order's last modified date.
+	 *
+	 * @since 4.6.0-dev
+	 *
+	 * @param \WC_Order $order order object
+	 * @param string $context if 'view' then the value will be filtered
+	 *
+	 * @return \WC_DateTime|null
+	 */
+	public static function get_date_modified( WC_Order $order, $context = 'edit' ) {
+
+		return self::get_date_prop( $order, 'modified', $context );
+	}
+
+
+	/**
+	 * Gets an order's paid date.
+	 *
+	 * @since 4.6.0-dev
+	 *
+	 * @param \WC_Order $order order object
+	 * @param string $context if 'view' then the value will be filtered
+	 *
+	 * @return \WC_DateTime|null
+	 */
+	public static function get_date_paid( WC_Order $order, $context = 'edit' ) {
+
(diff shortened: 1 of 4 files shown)


maxrice:
> Is the naming of the SV_WC_DateTime class too wonky?

I think this is the right approach, because we don't want to influence other code that may be checking for the existence of the `WC_DateTime` class.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -51,6 +51,116 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
...
+					$date->setTimezone( new DateTimeZone( wc_timezone_string() ) );
+
+				} catch ( Exception $e ) {
+
+					$date = null;
+				}
+			}
+		}
+
+		return $date;
```
maxrice: there's a possibility where `$date` is undefined here right? if `$order->prop` evaluates to `false`, then will `$date` have scope outside of that conditional?

PR #211, opened 2017-04-28
Fix: Hosted Gateway Authorizations

bekarice:
Ensures the hosted gateway class treats authorizations the same as the direct class, setting orders to "on hold"

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
@@ -446,7 +446,15 @@ abstract class SV_WC_Payment_Gateway_Hosted extends SV_WC_Payment_Gateway {
 			// If approved, payment is complete
 			if ( $response->transaction_approved() ) {
 
-				$order->payment_complete();
+				// determine whether we should complete payment or set to on-hold for later capture
+				if ( $this->supports( self::FEATURE_CREDIT_CARD_AUTHORIZATION ) && $this->perform_credit_card_authorization( $order ) ) {
+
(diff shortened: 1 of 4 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
```diff
@@ -446,7 +446,15 @@ protected function process_transaction_response( $order, $response ) {
 			// If approved, payment is complete
 			if ( $response->transaction_approved() ) {
 
-				$order->payment_complete();
+				// determine whether we should complete payment or set to on-hold for later capture
+				if ( $this->supports( self::FEATURE_CREDIT_CARD_AUTHORIZATION ) && $this->perform_credit_card_authorization( $order ) ) {
+
+					$this->mark_order_as_held( $order, $this->supports( self::FEATURE_CREDIT_CARD_AUTHORIZATION ) && $this->perform_credit_card_authorization( $order ) ? esc_html__( 'Authorization only transaction', 'woocommerce-plugin-framework' ) : $response->get_status_message(), $response );
```
ChaseWiseman: No need for the if/else here. It's identical to the `if ( $this->supports(...` above so it'll always return true if we got to this point.

Also, `esc_html__()` isn't needed for adding order notes, as that's done on output in the admin.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
```diff
@@ -446,7 +446,15 @@ protected function process_transaction_response( $order, $response ) {
...
+					SV_WC_Order_Compatibility::reduce_stock_levels( $order ); // reduce stock for held orders, but don't complete payment
+
+				} else {
```
ChaseWiseman: Just needs a following blank line :)

bekarice:
@ChaseWiseman done 🍻

PR #225, opened 2017-07-12
Feature: Apple Pay

ChaseWiseman:
This PR adds handlers to support Apple Pay on the checkout, cart, and single product pages.

## Gateway Implementation
1. Declare support for any gateway extending `SV_WC_Payment_Gateway` by using `self::FEATURE_APPLE_PAY`
2. Override the `SV_WC_Payment_Gateway::get_order_for_apple_pay()` method to add any data necessary for processing the transaction via the gateway's API. See `SV_WC_Payment_Gateway_Apple_Pay_Payment_Response` for returned data.

By default, the feature is inactive and will need to be activated by a filter:

`add_filter( 'wc_payment_gateway_{gateway_id}_activate_apple_pay', '__return_true' );`

This makes the **WooCommerce** -> **Settings** -> **Checkout** -> **Apple Pay** settings section available for configuration.

The classes and JS handle the entire Apple Pay flow, and when successful simply calls `SV_WC_Payment_Gateway::process_payment()`. From there the gateway handles everything like a normal transaction, and it's up to the gateway to set the appropriate API params with the data it sets in `SV_WC_Payment_Gateway::get_order_for_apple_pay()`.

## Limitations
* No subscription or pre-order (pay on release) transactions. If either exists in the cart, Apple Pay is unavailable.
* On the single product page, Apple Pay is only available for simple products. Additionally, the Apple Pay button is treated as a "Buy Now" button to quickly buy a single item, so the Quantity add-to-cart input is not a factor.
* Improvements are being logged in the GH project: https://github.com/skyverge/wc-plugin-framework/projects/3


bekarice:
@ChaseWiseman looks great, and SONP works well 🎉  really nice stuff here, the taxes and shipping handling in the Apple Pay modal is 💯 

One fix: the accepted currencies is failing, I think perhaps due to the supported / enabled Apple Pay gateway check: 

screen shot 2017-07-18 at 12 52 10 pm

As a final note, one thing we may want to disable Apple Pay on Membership-granting products like Subscriptions (not needed before deploy, we can follow with an update for AIM / CIM.

[The code here](https://github.com/skyverge/woocommerce-memberships/blob/master/includes/frontend/class-wc-memberships-checkout.php#L180) could likely benefit from a helper -- something like `wc_memberships_product_grants_membership_access( $product_id )` (bad name, I know) to return an array of plans the product grants access to, which could also be used within Memberships itself -- cc @unfulvio

That way we could bail on the product / cart / checkout. We could also consider making that Memberships `WC_Memberships_Checkout::force_registration()` check public so the FW could check that in the cart / checkout and bail if it's false. Something for the first feature release anyway.

PR #230, opened 2017-08-11
Improve TLS Notice

ChaseWiseman:
This removes some unreliable version checking, and softens the notice's impact so folks won't be as alarmed.

There is still the potential for false-positives, since the check for cURL version 7.34+ remains, but that's the best indicator I can come up with when deciding whether to show the notice.

Even with those cURL false-positives, the notice's new wording and appearance here should be a little more forgiving.

Closes #230

### DIFF
--- a/woocommerce/api/class-sv-wc-api-base.php
+++ b/woocommerce/api/class-sv-wc-api-base.php
@@ -805,7 +805,7 @@ abstract class SV_WC_API_Base {
 
 			$versions = curl_version();
 
-			// cURL is considered the minimum version that supports TLS 1.2
+			// cURL 7.34.0 is considered the minimum version that supports TLS 1.2
 			if ( version_compare( $versions['version'], '7.34.0', '<' ) ) {
 				$is_available = false;
 			}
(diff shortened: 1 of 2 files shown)


woocommerce/api/class-sv-wc-api-base.php
```diff
@@ -797,28 +797,29 @@ public function require_tls_1_2() {
...
+		// check the cURL version if installed
+		if ( is_callable( 'curl_version' ) ) {
 
-		$ssl_version = substr( $ssl_version, 0, -1 );
+			$versions = curl_version();
 
-		// if cURL and/or OpenSSL aren't up to the challenge, bail
-		if ( ! version_compare( $curl_version, '7.34.0', '>=' ) || ( 'OpenSSL' === $ssl_type && ! version_compare( $ssl_version, '1.0.1', '>=' ) ) ) {
-			return false;
+			// cURL is considered the minimum version that supports TLS 1.2
```
maxrice: typo? `// cURL 7.34.0 is considered...`

PR #236, opened 2017-09-08
Release v5: Base gateway tokenization

ChaseWiseman:
Move the main tokenization functionality to the base gateway class to allow hosted gateways to process subscriptions & pre-orders.


woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
```diff
@@ -102,7 +102,9 @@ public function process_payment( $order_id ) {
 			return array( 'result' => 'failure' );
 		}
 
-		WC()->cart->empty_cart();
+		if ( isset( WC()->cart ) ) {
```
maxrice: nitpick: adding an `instanceof` and `is_callable` check here might be a bit clearer, assuming the cart class name didn't change in 3.2

PR #240, opened 2017-09-11
Fix admin message handler to allow custom args

### LINKED ISSUE #233: Admin message handler does not respect params when showing messages
The use of [`wp_parse_args()` here](https://github.com/skyverge/wc-plugins/blob/pip-release-2242/woocommerce-pip/lib/skyverge/woocommerce/class-sv-wp-admin-message-handler.php#L246-L250) has its arguments reversed, so any params passed to `SV_WP_Admin_Message_Handler::show_messages()` will be overridden by the defaults.

ChaseWiseman:
Closes #233


unfulvio:
:ok:

PR #242, opened 2017-09-23
Some small suggestion for code improvements

blasvicco:
### API
_Replacing array() for []._
_Reducing code lines and removing not needed var assignation._
_Using Null coalescing operator._
_Reducing code lines._
_Creating new function is_get_or_head to avoid repeating code: in_array( strtoupper( $this->get_request_method() ), [ 'GET', 'HEAD' ] )._
_Adding new function get__request to avoid repeating code lines for get_request_body and get_sanitized_request_body functions._

### DIFF
--- a/woocommerce/api/abstract-sv-wc-api-json-request.php
+++ b/woocommerce/api/abstract-sv-wc-api-json-request.php
@@ -41,10 +41,10 @@ abstract class SV_WC_API_JSON_Request implements SV_WC_API_Request {
 	protected $path;
 
 	/** @var array The request parameters, if any */
-	protected $params = array();
+	protected $params = [];
 
 	/** @var array the request data */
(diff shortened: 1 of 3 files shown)


bekarice:
@blasvicco thanks for submitting this PR! We appreciate you taking the time to contribute here 😄 

I'm afraid we can't accept it at this time. In its present state, it requires PHP7+ due to the null coalescing operator usage. However, even without that, shortened array syntax was added in PHP 5.4, which we also don't require. The current branch requires PHP 5.2, and though we're [migrating to PHP 5.3+ now](https://github.com/skyverge/wc-plugin-framework/tree/release-v5), we don't plan to drop PHP 5.3 support until its usage is smaller among our own plugin users.

As such, I'm going to close this PR. Thanks again!

PR #245, opened 2017-10-04
Improve the PHP version notice wording and add date handling

ChaseWiseman:
Changed to v5.6 by May 2018.

@bekarice mind taking a look at the wording here? Before May 1 next year:

php-notice

After:

php-notice-after

The notice is also now dismissible except on the plugin settings page if there is one.


bekarice:
@ChaseWiseman looks great to me, nice work!

PR #259, opened 2018-01-12
v4: WC 3.3 compat

ChaseWiseman:


ChaseWiseman:
@ragulka went ahead and merged this so folks with Mondays starting before mine aren't held up from starting compatibility 😄

No rush, if there are any review changes needed I'll make them directly on the `legacy-v4` branch.

Issue #190, opened 2017-01-03
Add Link to Download Minimum WooCommerce Version


bekarice:
Add a link to the "WooCommerce" outdated notice to directly download our minimum required WC version from wordpress.org.

In this case, it will likely be `https://downloads.wordpress.org/plugin/woocommerce.2.5.5.zip`

bekarice:
closed via 8bd53d5893beb968336bee9614a4d5ca666f19fb

PR #260, opened 2018-01-12
[#190] Add WC minimum download link to the outdated notice

ChaseWiseman:


ChaseWiseman:
@bekarice decided to sneak this in since we'll be doing a rare all-plugin deploy, and because of the bootstrap's limitations it'll be the only time to guarantee it's displayed. lmk if it looks okay to you!

notice

bekarice:
@ChaseWiseman good idea, that looks great to me

PR #263, opened 2018-01-30
5.1: Plugin review prompts

### LINKED ISSUE #258: Add Plugin Review Links
Consider addition of a "Review" link for plugin action links using the `{plugin_url}#comments` link and other lifecycle prompts for reviews and cross-sells.

ChaseWiseman:
- [x] Allow plugins to define their own sales/review page URLs
- [x] Add a "Review" link to plugin action links
- [x] Allow easy registration of plugin milestone events and display review prompts when needed
- [x] Add standard gateway milestones like first payment and first refund
- [x] Detect upgrades so milestone messages are only displayed for first-time installs
- [x] Determine if we should limit the number of messages to display, i.e. if the user has dismissed x number of review prompts, don't add any more for later milestones

The approach I took here is to only display milestones for new installs. This prevents existing users from seeing "first payment!" notices if they've been processing payments for a long time.

Plugins can call `Lifecycle::trigger_milestone( $id, $message, $since );` to register a notice when a milestone is reached. They can also omit `$since` for milestones that have existed in the plugin all along, like refunds in gateways, so that those messages are only displayed to brand new users. We don't have a great way to detect new vs. existing users, so we consider [all upgrades](https://github.com/skyverge/wc-plugin-framework/blob/5.1-plugin-review-prompts-258/woocommerce/Lifecycle.php#L86-L92) as existing users. This may leave out a subset of users who for some reason upgrade the plugin without having ever completed early milestones, but I think that's an okay compromise to make sure we don't bug existing users.

Otherwise, specifying `$since` can be used for new features so that even existing users get the notice. So if Memberships adds a totally new export feature in v1.11, we can pass `$since = '1.11.0'` and all users will see that when it first happens.

We also limit the number of milestone notices that can be added to 3 so that users aren't bombarded with review prompts if a plugin has a lot of milestones.

Closes #258

### DIFF
--- a/woocommerce/Lifecycle.php
+++ b/woocommerce/Lifecycle.php
@@ -39,6 +39,9 @@ if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_0_1\\Plugin\\
 class Lifecycle {
 
 
+	/** @var string minimum milestone version */
+	private $milestone_version;
+
 	/** @var \SkyVerge\WooCommerce\PluginFramework\v5_0_1\SV_WC_Plugin plugin instance */
 	private $plugin;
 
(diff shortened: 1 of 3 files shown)


woocommerce/Lifecycle.php
```diff
@@ -0,0 +1,292 @@
...
+	 * triggering notices for milestones that have long passed, like a payment
+	 * gateway's first successful payment. Omitting $since will assume the
+	 * milestone has always existed and should only trigger for fresh installs.
+	 *
+	 * @since 5.1.0-dev
+	 *
+	 * @param string $id milestone ID
+	 * @param string $message message to display to the user
+	 * @param string $since the version since this milestone has existed in the plugin
+	 */
```
unfulvio: missing return tag - this method seems to return false on version compare failure or void if sanity check passes

should it return something or bool? if nothing, can be just `return;`

woocommerce/Lifecycle.php
```diff
@@ -0,0 +1,292 @@
...
+				__( 'Cowabunga', 'woocommerce-plugin-framework' ),
+				__( 'Congratulations', 'woocommerce-plugin-framework' ),
+				__( 'Hot dog', 'woocommerce-plugin-framework' ),
+			);
+
+			$message = $exclamations[ array_rand( $exclamations ) ] . ', ' . esc_html( $custom_message ) . ' ';
+
+			$message .= sprintf(
+				/* translators: Placeholders: %1$s - plugin name, %2$s - <a> tag, %3$s - </a> tag, %4$s - <a> tag, %5$s - </a> tag */
+				__( 'Are you having a great experience with %1$s so far? Please consider %2$sleaving a review%3$s! If things aren\'t going quite as expected, we\'re happy to help -- please %4$sreach out to our support team%5$s.', 'woocommerce-plugin-framework' ),
```
unfulvio: consider wrapping this message in double quotes instead of escaping apostrophes

PR #266, opened 2018-02-09
5.1: New Payment Methods UI

### LINKED ISSUE #178: Update Payment Method UI
It would be cool to add a basic UI for updating credit card expiration dates from My Account. Gateways like Authorize.Net CIM now support this via their API, and there are probably more to come.

It shouldn't be too involved to add an "Edit" action button next to "Delete" that reveals some expiration date inputs or something similar. Then we could add a new feature flag and some stub methods to handle everything else.

At checkout, this would also allow us to adjust the way expired cards are shown. Rather than showing the expiration date: `(expires mm/yy)` we can show `(expired - update)`, linking the "update" text to to our new update endpoint (or perhaps doing it inline, not sure if this would be possible yet). Highlighting / making the expired text red would also help draw attention to this.

As a final level of polish, if this is possible, we could disable placing an order for an expired method, but not sure if that's too obtrusive.

ChaseWiseman:
Adds "nicknames" for payment methods, as well as front-end editing capability.

Closes #178


woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-my-payment-methods.coffee
```diff
@@ -0,0 +1,187 @@
...
+			button.siblings( '.save-payment-method' ).hide()
+			button.siblings( '.delete-payment-method' ).show()
+
+			$( ".wc-#{@slug}-my-payment-methods" ).removeClass( 'editing' )
+
+
+		# Blocks the payment methods table UI.
+		#
+		# @since 5.1.0-dev
+		block_ui: -> $( ".wc-#{@slug}-my-payment-methods" ).parent( 'div' ).block( message: null, overlayCSS: background: '#fff',opacity: 0.6 )
```
maxrice: space before `opacity`

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -799,6 +1043,13 @@ protected function supports_add_payment_method() {
 	}
 
 
+	protected function is_payment_methods_page() {
```
maxrice: does this method need a doc block?

PR #276, opened 2018-03-30
Fix a PHP warning in 7.2 from the gateway settings

ChaseWiseman:
Also cleans the logic up a bit and accounts for a single paid status.

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -3141,7 +3141,8 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 
 			$last_status = array_pop( $paid_statuses );
 
-			array_push( $paid_statuses, "or {$last_status}" );
+			/* translators: a conjuction used in the list of order status names, such as "Processing, Completed, or Shipped" */
+			array_push( $paid_statuses, __( 'or', 'woocommerce-plugin-framework' ) . ' ' . $last_status );
 		}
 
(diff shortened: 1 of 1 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -3134,19 +3134,24 @@ protected function add_authorization_charge_form_fields( $form_fields ) {
...
 
 		// do some oxford comma magic
-		$last_status = array_pop( $paid_statuses );
-		array_push( $paid_statuses, "or {$last_status}" );
-		$separator = count( $last_status ) < 3 ? ' ' : ', ';
+		if ( count( $paid_statuses ) > 1 ) {
+
+			$last_status = array_pop( $paid_statuses );
+
+			array_push( $paid_statuses, "or {$last_status}" );
```
unfulvio: This doesn't seem to be translatable though.

By the way Memberships had since v1.0.0 an utility function called `wc_memberships_list_items()` to do what you're doing here, but makes the 'or' conjunction customizable (e.g. use `and` instead of `or`) and translatable.

You might consider importing the function `wc_memberships_list_items()` in framework as I've seen a few uses of that in WC plugins.

bekarice: I'd stick with this format and make a FW helper if anything, as the one in Memberships doesn't use an Oxford comma though it should 😸then migrate any other usages to the FW helper

ChaseWiseman: added an issue for the helper: https://github.com/skyverge/wc-plugin-framework/issues/279

PR #287, opened 2018-05-14
Target the form row parent when moving the CSC input

### LINKED ISSUE #284: Target parent <p> when handling saved payment methods
Coming in WC 3.4, the payment form inputs will be wrapped in a `<span>` to facilitate [this description display enhancement](https://github.com/woocommerce/woocommerce/commit/3bf567279b6bc6058505067d2d8eeece999e77af).

This causes a problem with our payment form JS when toggling between saved methods & the new method form while moving the CSC field - we [target the parent](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee#L302) but that will now be the span instead of the real row parent. We should specifically target `<p>` (or the form row class) so the full form row gets moved.

ChaseWiseman:
Closes #284

### DIFF
--- a/woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
+++ b/woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
@@ -299,7 +299,7 @@ jQuery( document ).ready ($) ->
 			csc_required_for_tokens  = @csc_required_for_tokens
 
 			$new_payment_method_selection = $( "div.js-wc-#{ id_dasherized }-new-payment-method-form" )
-			$csc_field = $new_payment_method_selection.find( '.js-sv-wc-payment-gateway-credit-card-form-csc' ).parents( '.form-row' )
+			$csc_field = $new_payment_method_selection.find( '.js-sv-wc-payment-gateway-credit-card-form-csc' ).closest( '.form-row' )
 
 			# show/hide the saved payment methods when a saved payment method is de-selected/selected
 			$( "input.js-wc-#{ @id_dasherized }-payment-token" ).change ->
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
```diff
@@ -299,10 +299,11 @@ jQuery( document ).ready ($) ->
 			csc_required_for_tokens  = @csc_required_for_tokens
 
 			$new_payment_method_selection = $( "div.js-wc-#{ id_dasherized }-new-payment-method-form" )
-			$csc_field = $new_payment_method_selection.find( '.js-sv-wc-payment-gateway-credit-card-form-csc' ).parent()
+			$csc_field = $new_payment_method_selection.find( '.js-sv-wc-payment-gateway-credit-card-form-csc' ).parents( '.form-row' )
```
ragulka: Consider using `closest()` instead of `parents()`, as it will only match the first ancestor, while the latter will match all, which 1) could cause unintended effects on borked HTML layouts 2) be a tad slower

See https://stackoverflow.com/a/10499468

woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
```diff
@@ -322,7 +323,7 @@ jQuery( document ).ready ($) ->
 					# move the CSC field back into its regular spot
 					if csc_required_for_tokens
 						$csc_field.removeClass( 'form-row-first' ).addClass( 'form-row-last' )
-						$new_payment_method_selection.find( '.js-sv-wc-payment-gateway-credit-card-form-expiry' ).parent().after( $csc_field )
+						$new_payment_method_selection.find( '.js-sv-wc-payment-gateway-credit-card-form-expiry' ).parents( '.form-row' ).after( $csc_field )
```
ragulka: Same as above, prefer `closest()` instead

PR #288, opened 2018-05-14
Add upgrade warning for WC installs < 3.0

### LINKED ISSUE #282: Add an admin notice for WC 2.6 users that support will soon be dropped
We've got a fair number of users still running WC 2.6, so we should add an admin notice to encourage folks to upgrade. There are enough using it that we can't yet drop support but we will eventually, likely with WC 4.0 compatibility, so we should start notifying users early about upgrading.

For now this can be a regular notice (`.notice-info`), then later in the year we can escalate it to a warning (`.notice-warning`).

ChaseWiseman:
Closes #282

### DIFF
--- a/woocommerce/class-sv-wc-plugin.php
+++ b/woocommerce/class-sv-wc-plugin.php
@@ -517,7 +517,7 @@ abstract class SV_WC_Plugin {
 
 			$message = sprintf(
 				/* translators: Placeholders: %1$s - WooCommerce version number, %2$s - <strong>, %3$s - </strong>, %4$s - Plugin name, %5$s - <a> tag, %6$s - </a> tag */
-				__( 'Hey there! We\'ve noticed that your site is running version %1$s of WooCommerce. While this version is still supported, %2$sWooCommerce 3.0 or higher will soon be required%3$s by %4$s. We recommend you %5$supdate WooCommerce%6$s to the latest version as soon as possible.', 'woocommerce-plugin-framework' ),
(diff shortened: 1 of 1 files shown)


ChaseWiseman:
@bekarice would love your messaging expertise here when you get a chance, thanks!

woocommerce/class-sv-wc-plugin.php
```diff
@@ -511,6 +511,23 @@ protected function add_dependencies_admin_notices() {
 
 			$sv_wc_php_notice_added = true;
 		}
+
+		// display a notice that WC < 3.0 support will soon be dropped
+		if ( 'wc-settings' === SV_WC_Helper::get_request( 'page' ) && SV_WC_Plugin_Compatibility::is_wc_version_lt( '3.0' ) ) {
+
+			$message = sprintf(
+				/* translators: Placeholders: %1$s - WooCommerce version number, %2$s - <strong>, %3$s - </strong>, %4$s - Plugin name, %5$s - <a> tag, %6$s - </a> tag */
+				__( 'Hey there! We\'ve noticed that your site is running version %1$s of WooCommerce. While this version is still supported, %2$sWooCommerce 3.0 or higher will soon be required%3$s by %4$s. We recommend you %5$supdate WooCommerce%6$s to the latest version as soon as possible.', 'woocommerce-plugin-framework' ),
```
bekarice: should we use double quotes instead so we don't have to escape the apostrophe here?

bekarice: Let's simplify it just a tad maybe?

```
Hey there! We've noticed that your site is running version %1$s of WooCommerce, but %2$sWooCommerce 3.0 or higher will soon be required%3$s by %4$s.
```

ChaseWiseman: Wording tweaked in 2cda8181396de3b868e3932121eeea3936d8dbd6 👍 

We can't use double quotes for that string though as the `sprintf` placeholders won't parse correctly

PR #302, opened 2018-06-26
Add SV_WC_Helper:: list_array_items()

### LINKED ISSUE #279: Add a "list items" helper method
Make it easy to print a formatted list from an array of strings. Like the list function [from Memberships](https://github.com/skyverge/woocommerce-memberships/blob/master/includes/functions/wc-memberships-functions-misc.php#L28-L46), but with more Oxford comma 😄

Can be adapted from and replace the gateway [paid order capture setting](https://github.com/skyverge/wc-plugin-framework/blob/fix-gateways-settings-warning/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php#L3136-L3148).

ChaseWiseman:
Closes #279

### DIFF
--- a/woocommerce/class-sv-wc-helper.php
+++ b/woocommerce/class-sv-wc-helper.php
@@ -388,6 +388,44 @@ if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_1_4\\SV_WC_He
 		}
 
 
+		/**
+		 * Lists an array as text.
+		 *
+		 * Takes an array and returns a list like "one, two, three, and four"
+		 * with a (mandatory) oxford comma.
+		 *
+		 * @since 5.2.0-dev
+		 *
+		 * @param array $items items to list
+		 * @param string|null $conjunction coordinating conjunction, like "or" or "and"
+		 * @param string $separator list separator, like a comma
+		 * @return string
+		 */
+		public static function list_array_items( array $items, $conjunction = null, $separator = '' ) {
+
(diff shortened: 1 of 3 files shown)


woocommerce/class-sv-wc-helper.php
```diff
@@ -388,6 +388,44 @@ public static function array_to_xml( $xml_writer, $element_key, $element_value =
...
+
+			// append the conjunction to the last item
+			if ( count( $items ) > 1 ) {
+
+				$last_item = array_pop( $items );
+
+				array_push( $items, trim( "{$conjunction} {$last_item}" ) );
+			}
+
+			// only use a comma if needed and no separator was passed
```
unfulvio: I see this is essentially a re-implementation of a function that was present in Memberships and I guess other plugins. The problem I see here, especially with the handling of the conjunction and the different handling of 2 vs more than 2 items is that this might not be 100% translation proof. But I don't have a better alternative save for just listing all the items comma separated. That being said we haven't encountered yet a problem with this function; if we do I suppose we can address it in this FW method in the future.

Also you could the logic below determining a separator within `count( $items ) > 1` condition, and leave the separator to its default empty string, given that a separator won't be used for lists of just one item.

ChaseWiseman: yeah I tried to account for translation as much as possible, which may not be 100% given the nature of a dynamically generated list like this. We can certainly adjust things based on feedback in the future.

PR #316, opened 2018-08-30
[#286] Add debug log notice in production mode

### LINKED ISSUE #286: Add an admin notice when gateway debug logging is enabled
Alert merchants to the fact that debug logging is enabled, and that unless they are actively working on debugging a problem it should be disabled. This will help prevent users from leaving logging on well after they've completed troubleshooting an issue with the support team, and unnecessarily logging transaction data.

* This should be dismissible, but always on plugin settings
* Ensure the setting description (or tooltip) itself conveys the same idea

ChaseWiseman:
Message:

>Heads up! <gateway-title> is currently configured to log transaction data for debugging purposes. If you are not experiencing any problems with payment processing, we recommend [turning off Debug Mode](#)

Closes #286


woocommerce/payment-gateway/class-sv-wc-payment-gateway-plugin.php
```diff
@@ -662,6 +665,35 @@ protected function add_currency_admin_notices() {
...
+	protected function add_debug_setting_notices() {
+
+		foreach ( $this->get_gateways() as $gateway ) {
+
+			if ( $gateway->is_enabled() && $gateway->is_production_environment() && ! $gateway->debug_off() ) {
+
+				$is_gateway_settings = $this->is_payment_gateway_configuration_page( $gateway->get_id() );
+
+				$message = sprintf(
+					__( 'Heads up! %1$s is currently configured to log transaction data for debugging purposes. If you are not experiencing any problems with payment processing, we recommend %2$sturning off Debug Mode%3$s', 'woocommerce-plugin-framework' ),
```
jdeeburke: Does this need translator notes?

Issue #317, opened 2018-08-31
Regression in payment capture alert


bekarice:
Payment capture alerts should not use `wc_price()` to format capture amount, or should avoid escaping this HTML. Best guess is a regression due to #309, so we should strip tags first.

![screen shot 2018-08-31 at 5 15 54 pm](https://user-images.githubusercontent.com/4983547/44936733-e61eea80-ad43-11e8-8ae8-df0fdf826a6a.png)

unfulvio:
:melon:

Issue #313, opened 2018-08-20
The capture UI is displayed when editing a subscription


ChaseWiseman:
We need to add some checks to make sure the updated capture UI is only displayed on the Edit Order screen. As it is now, it also shows up for subscriptions.

Issue #300, opened 2018-06-21
Orders updated via the API do not get auto-captured


tamarazuk:
The new auto-capturing feature doesn't work when an order's status is updated via the REST API because the `SV_WC_Payment_Gateway_Admin_Order` class that handles this [is only loaded in the admin](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/class-sv-wc-payment-gateway-plugin.php#L282-L291)

Issue #285, opened 2018-05-08
Capturing charges doesn't work when multiple v5 gateways are running


tamarazuk:
If two or more v5 payment gateways are running, it's impossible to capture an authorized charge: 

![](http://cloud.skyver.ge/1a1v2e1m2U1Q/Screen%20Recording%202018-05-07%20at%2009.32%20PM.gif)

Inspecting the AJAX request, I noticed the action name was referencing Realex HPP while the payment was made with Auth.net AIM

![](http://cloud.skyver.ge/3X2c090G2x3Y/[95c8f190473ffec7ab08549f6c4c4888]_Screen%20Shot%202018-05-07%20at%209.22.58%20PM.png)

PR #320, opened 2018-09-07
Gateway: Capture tweaks & fixes

ChaseWiseman:
Moves most of the capture code into its own handler to be made available everywhere and further clean up the main gateway class.

- [x] #317 
- [x] #313
- [x] #300 
- [x] #285

### DIFF
--- a/woocommerce/payment-gateway/Handlers/Capture.php
+++ b/woocommerce/payment-gateway/Handlers/Capture.php
@@ -1,0 +1,424 @@
+<?php
+/**
+ * WooCommerce Payment Gateway Framework
+ *
+ * This source file is subject to the GNU General Public License v3.0
+ * that is bundled with this package in the file license.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://www.gnu.org/licenses/gpl-3.0.html
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@skyverge.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade the plugin to newer
+ * versions in the future. If you wish to customize the plugin for your
+ * needs please refer to http://www.skyverge.com
+ *
+ * @package   SkyVerge/WooCommerce/Payment-Gateway/Admin
+ * @author    SkyVerge
+ * @copyright Copyright (c) 2013-2018, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_2_2\Payment_Gateway\Handlers;
+
+use SkyVerge\WooCommerce\PluginFramework\v5_2_2 as Framework;
+
+defined( 'ABSPATH' ) or exit;
+
+if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_2_2\\Payment_Gateway\\Handlers\\Capture' ) ) :
+
+/**
+ * The transaction capture handler.
+ *
+ * @since 5.3.0-dev
+ */
+class Capture {
+
+
+	/** @var Framework\SV_WC_Payment_Gateway payment gateway instance */
+	private $gateway;
+
+
+	/**
+	 * Capture constructor.
+	 *
+	 * @since 5.3.0-dev
+	 *
+	 * @param Framework\SV_WC_Payment_Gateway $gateway payment gateway instance
+	 */
+	public function __construct( Framework\SV_WC_Payment_Gateway $gateway ) {
+
+		$this->gateway = $gateway;
+
+		// auto-capture on order status change if enabled
+		if ( $gateway->supports_credit_card_capture() && $gateway->is_paid_capture_enabled() ) {
+			add_action( 'woocommerce_order_status_changed', array( $this, 'maybe_capture_paid_order' ), 10, 3 );
+		}
+	}
+
+
+	/**
+	 * Captures an order on status change to a "paid" status.
+	 *
+	 * @internal
+	 *
+	 * @since 5.3.0-dev
+	 *
+	 * @param int $order_id order ID
+	 * @param string $old_status status being changed
+	 * @param string $new_status new order status
+	 */
+	public function maybe_capture_paid_order( $order_id, $old_status, $new_status ) {
+
+		$paid_statuses = Framework\SV_WC_Plugin_Compatibility::wc_get_is_paid_statuses();
+
+		// bail if changing to a non-paid status or from a paid status
+		if ( ! in_array( $new_status, $paid_statuses, true ) || in_array( $old_status, $paid_statuses, true ) ) {
+			return;
+		}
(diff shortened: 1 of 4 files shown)


woocommerce/payment-gateway/Handlers/Capture.php
```diff
@@ -0,0 +1,424 @@
...
+ * versions in the future. If you wish to customize the plugin for your
+ * needs please refer to http://www.skyverge.com
+ *
+ * @package   SkyVerge/WooCommerce/Payment-Gateway/Admin
+ * @author    SkyVerge
+ * @copyright Copyright (c) 2013-2018, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_2_2\Payment_Gateway\Handlers;
```
unfulvio: is this supposed to be tagged `5.3.0` as the methods point out in their `since` tags and v5.2.2 is released already? or is this for 5.2.3?

woocommerce/payment-gateway/Handlers/Capture.php
```diff
@@ -0,0 +1,424 @@
...
+
+	/**
+	 * Lets gateways handle any specific capture failure results for the order.
+	 *
+	 * @since 5.3.0-dev
+	 *
+	 * @param \WC_Order $order WooCommerce order object
+	 * @param Framework\SV_WC_Payment_Gateway_API_Response $response API response object
+	 */
+	public function do_capture_failed( \WC_Order $order, Framework\SV_WC_Payment_Gateway_API_Response $response ) { }
```
unfulvio: I suppose this can't/shouldn't be an abstract child classes must implement?

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php
```diff
@@ -81,31 +78,54 @@ public function __construct( SV_WC_Payment_Gateway_Plugin $plugin ) {
...
 
 
+	/**
+	 * Enqueues the assets for the Edit Order screen.
+	 *
+	 * @since 5.3.0-dev
+	 *
+	 * @param \WC_Order $order order object
+	 */
+	protected function enqueue_edit_order_assets( \WC_Order $order ) {
```
unfulvio: should this be private? or is it expected individual gateways to extend it?

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php
```diff
@@ -81,31 +78,54 @@ public function __construct( SV_WC_Payment_Gateway_Plugin $plugin ) {
 	 */
 	public function enqueue_scripts( $hook_suffix ) {
 
-		if ( 'shop_order' === get_post_type() && 'post.php' === $hook_suffix ) {
+		// Order screen assets
+		if ( 'shop_order' === get_post_type() ) {
 
-			wp_enqueue_script( 'sv-wc-payment-gateway-admin-order', $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/js/admin/sv-wc-payment-gateway-admin-order.min.js', array( 'jquery' ), SV_WC_Plugin::VERSION, true );
+			// Edit Order screen assets
+			if ( 'post.php' === $hook_suffix ) {
```
unfulvio: this condition can be merged in the previous statement as it was before, no?

unfulvio: You can write each concatenated `&&` condition in a new line inside a single if statement - it makes clear from the beginning there are multiple requirements for a condition. If this is moved below I might assume there's then an if/else branch, at a glance.

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php
```diff
@@ -197,7 +217,9 @@ public function process_capture_charge_bulk_order_action() {
 
 				$order = wc_get_order( $order_id );
 
-				$this->maybe_capture_charge( $order );
+				if ( $order && $gateway = $this->get_order_gateway( $order ) ) {
```
unfulvio: We don't have a rule in our code style guide about this.

There are some sources that discourage having a variable assignment in condition. It may be confusing or prone to errors (maybe you want to check if a variable exists or is of a type and write `=` instead of `==` and vice versa.

I personally don't think it's an issue, when you have 1-2 conditions only. If there's more than 2 maybe it's best to break down the `if` statement into multiple statements.

However, regardless of this, maybe it's a good idea to write the condition where you also assign a variable in parenthesis like so:

```php
if ( $order && ( $gateway = $this->get_order_gateway( $order ) ) ) {
	// ...
}
```

This should highlight that you're assigning (or overwriting) a variable. I personally do this, but we can discuss in WC dev meeting what's best here, or leave it to personal preferencee.

woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -1740,131 +1746,46 @@ protected function complete_payment( \WC_Order $order, SV_WC_Payment_Gateway_API
...
-	 *
-	 * $order->capture->amount
-	 * $order->capture->description
+	 * This was intentionally not moved to the capture handler since we'll likely be refactoring how this information is
+	 * set in the future, and plenty of gateways override it.
 	 *
 	 * @since 4.5.0
 	 *
 	 * @param \WC_Order|int $order the order being processed
 	 * @param float $amount amount to capture
```
unfulvio: specify it can be omitted - `null` type

woocommerce/payment-gateway/Handlers/Capture.php
ChaseWiseman: I'm keeping the namespace change for last when we actually tag this release, so plugins can more easily test these things without having to change their namespace too

woocommerce/payment-gateway/Handlers/Capture.php
ChaseWiseman: nah, most plugins won't have any use for it but it still needs to exist just in case the plugin needs to do any special handling based on the API response code or something

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php
ChaseWiseman: yes it's possible that some plugin may want to extend it

woocommerce/payment-gateway/Handlers/Capture.php
unfulvio: cool

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php
ChaseWiseman: I'm not a fan of the multi-line if statements generally, and in this case I'm preferring to keep each statement separate since these are conditions for when certain scripts are enqueued.

The first checking for the general order post type, then specifically the edit screen, etc... which makes it easy to add other enqueue methods that are meant for different screens without refactoring the conditions.

PR #322, opened 2018-09-18
Setup Wizard: Connection Handler

unfulvio:
WIP - do not merge

Working on a Connection Handler and Setup Wizard update in order to allow plugins that connect to external services (MailChimp, Google, Gateways...) to set up their connection to such services in the Setup Wizard context in a dedicated step.

The plugins will need to abstract their connection to a shared `Connection_Handler` class which will offer known methods to connect, disconnect and retrieve the connection status of the plugin in regards of the external service.

The Setup Wizard will use these method to build its UI to determine whether a connection failed or succeeded and move onto the next step(s).

 * **See:** [Initial proposal](https://docs.google.com/document/d/15DkfTtW8Xf7dJDN78_rG2DuARYNNg_w1K1i3E3MCUkQ/edit) (but this is evolving rapidly in its implementation)
 * **See:** https://github.com/skyverge/wc-memberships-addons/pull/52#issuecomment-422985743 an interesting conversation as we got hands on the first implementation with MailChimp Sync
 * **Note:** We decided not to build any specific UI for now and focus only on the connection handler and the step implementation. The shared UI components will be in the form of known CSS classes that a shared AJAX callback may be able to use to handle the connection attempt and populate result status.


PierrishMoore87:
I hope it work

ChaseWiseman:
Closing as no longer relevant

Issue #331, opened 2019-01-08
Update the plugin loader sample to bring in line with code standards and latest features


jdeeburke:
### PR in #332 

The [plugin loader sample](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/woocommerce-framework-plugin-loader-sample.php) needs some tweaking for best practices and to implement new functionality in the framework.

PR #332, opened 2019-01-08
Update Plugin Loader sample class

jdeeburke:
see #331


jdeeburke:
Hey @ChaseWiseman couple of tweaks here, some of which identified by @unfulvio in the CAO refactor review.

Also added the stub out for autoloading, curious if there are any other new features that should be added here that I'm forgetting.

PR #338, opened 2019-07-23
[ch12281] Add unsupported WC core admin notices 

unfulvio:
Adds notices if the merchant is running a WooCommerce version 2 minor releases _behind_ the latest available version on WordPress.org.

### Details

Individual plugins can override this criteria by passing a `latest_wc_versions` argument inside the `$args` param passed to `SV_WC_Plugin` constructor.

This argument can be 0 or empty, or can be a float. When float, for example the default `0.2` it means the framework will treat this as an inverse partial semver and assume x.y versions behind the latest. In this case (default) value, 2 minor versions behind latest are supported. If a plugin wants to offer longer support up to three versions this could be `0.3`. 

Since the Framework is an open source project that may be used by third parties I thought it would be most considerate to offer this kind of flexibility. This may be helpful for ourselves (should we have a different handling or no handling for specific plugins) but also for third parties with a different support policy (for example `1.0` will offer support up to a major version behind).

To know which is the true last version of WooCommerce, the framework will issue a request to WordPress.org via WordPress.org API. This will fetch all information about the `woocommerce` plugin including its versions. `trunk`, release candidates, betas, etc. will be removed from the versions array. The last element in the array is the last available version. From this version, the framework loops back until it cycles the specified parameter (default `0.2`). Since semver isn't linear, it does so in a way that doesn't assume in advance what "2 versions behind" means, given this could be 4.0.0, and two minor versions behind could mean 3.8.9 and 3.8.8 OR 3.9.2 and 3.9.1 as well. 

To avoid querying the WordPress.Org API too often, the latest WooCommerce versions are stored in a weekly transient that can be shared by all frameworked plugins.

Finally, if the installed version is 1-2 (or specified) versions behind, it will display a notice reminding the merchant about the imminent support drop. If the version is even older, it will issue a warning notice telling the merchant the current WooCommerce version is no longer supported by the plugin that issued the notice. Once the merchant updates, the notice is no longer shown.

## QA

- [ ] To test an implementation, have a look at this [Memberships PR](https://github.com/skyverge/woocommerce-memberships/pull/727#issue-300206508). To trigger the notice you can simply edit the `WC_VERSION` constant from the active WooCommerce plugin. 
- [ ] If the constant is moved within 2 minor versions behind, a yellow notice should appear.
- [ ] If the constant is moved further behind, a red warning notice should appear.
- [ ] Either notice should prompt the user to update WooCommerce, linking to the WordPress update page.

### DIFF
--- a/woocommerce/class-sv-wc-admin-notice-handler.php
+++ b/woocommerce/class-sv-wc-admin-notice-handler.php
@@ -54,21 +54,23 @@ class SV_WC_Admin_Notice_Handler {
 
 
 	/**
-	 * Initialize and setup the Admin Notice Handler
+	 * Initializes and sets up the admin notices handler.
 	 *
 	 * @since 3.0.0
+	 *
+	 * @param SV_WC_Plugin main instance
 	 */
 	public function __construct( $plugin ) {
 
-		$this->plugin      = $plugin;
+		$this->plugin = $plugin;
 
 		// render any admin notices, delayed notices, and
-		add_action( 'admin_notices', array( $this, 'render_admin_notices'         ), 15 );
-		add_action( 'admin_footer',  array( $this, 'render_delayed_admin_notices' ), 15 );
-		add_action( 'admin_footer',  array( $this, 'render_admin_notice_js'       ), 20 );
+		add_action( 'admin_notices', [ $this, 'render_admin_notices'         ], 15 );
+		add_action( 'admin_footer',  [ $this, 'render_delayed_admin_notices' ], 15 );
+		add_action( 'admin_footer',  [ $this, 'render_admin_notice_js'       ], 20 );
 
 		// AJAX handler to dismiss any warning/error notices
-		add_action( 'wp_ajax_wc_plugin_framework_' . $this->get_plugin()->get_id() . '_dismiss_notice', array( $this, 'handle_dismiss_notice' ) );
+		add_action( 'wp_ajax_wc_plugin_framework_' . $this->get_plugin()->get_id() . '_dismiss_notice', [ $this, 'handle_dismiss_notice' ] );
 	}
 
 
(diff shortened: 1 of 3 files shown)


ChaseWiseman:
@unfulvio nice writeup! The branch is working well for me.

I'm wondering about the `latest_wc_versions` param and how it's used, as it isn't super clear from its naming.

Could we simplify things a bit by having that param represent the lowest supported version? So if a plugin _needs_ to keep support for a version that's outside the official "current minus 2" standard, they could set `'lowest_supported_wc_version' => '3.2'` to override.

In that case, I think a getter could house the other logic like pulling the versions via API if that's not overridden, and in the notice handling we'd only need to check `$installed_version <= $this->get_lowest_supported_wc_version()` to display the notice. What do you think?

unfulvio:
@ChaseWiseman added some change in the most recent commits.

I've also included a clause for what we spoke about in our last 1:1. If the current published version of WC is 3.6.x or 3.7.x then just mention about a future WC support policy change. No need to compare dates or anything. We can drop this later in a future FW version when 3.8 is the latest version available.

> I'm wondering about the latest_wc_versions param and how it's used, as it isn't super clear from its naming.

I've renamed the property into `min_wc_semver`.

> Could we simplify things a bit by having that param represent the lowest supported version? So if a plugin needs to keep support for a version that's outside the official "current minus 2" standard, they could set 'lowest_supported_wc_version' => '3.2' to override.

But then what's the point of having a notice if anything before a fixed version isn't supported? Can't the plugin just include that in the plugin loader's `MINIMUM_WC_VERSION` constant? If that's so, it would just behave as now. Then, they can set that `min_wc_semver` property in the main class to `null` or empty and the notices comparing last 2 semver won't get added. 

> In that case, I think a getter could house the other logic like pulling the versions via API if that's not overridden, and in the notice handling we'd only need to check $installed_version <= $this->get_lowest_supported_wc_version() to display the notice. What do you think?

A comparison needs to return different values to show a different content if they are running exactly 2 versions behind or beyond that. `version_compare` returns exactly what we need. But the notices also need some information (last available version) that would be contained in the logic that pulls the latest version published).

If plugins are overriding `add_admin_notices()` without calling `parent::add_admin_notices()`, these notices won't be triggered.

PR #339, opened 2019-07-24
[ch12277] Add support to WooCommerce Admin Notes

unfulvio:
## Add support for WooCommerce Admin Notes

Replace/integrate our WordPress notices handler to display notices, warnings and messages to merchants in admin with admin notes provided by the next generation WooCommerce Admin interface.

### Premise 

If a merchant has [WooCommerce Admin](https://wordpress.org/plugins/woocommerce-admin/) feature plugin installed, or later on when this plugin gets merged into WooCommerce core, there's a new place where important messages exclusive to WooCommerce are displayed, without getting mixed up with other traditional WordPress admin notices from other plugins.

In WooCommerce Admin, such admin messages are normalized as Admin Notes. These are abstracted into `WC_Admin_Note` objects which is yet another WooCommerce object collection built on top of the `WC_Data_Store` model (stored in a dedicated `wpdb` table) and made accessible to a React UI via WC REST API. Such notes, being db objects, carry a lot more information and logic than WordPress notices could (in fact, WordPress notices aren't standardized into any store: it's just a common CSS/HTML structure that could be attached to a known action hook). 

Since there will be a grace time where both installations without WooCommerce Admin and with WooCommerce Admin will be supported by our plugins, we need to account for both behaviors and continue display traditional WordPress notices where WooCommerce Admin isn't available. 

As such, we can leverage our current admin notice handler to register and display a traditional WordPress notice OR create (once, as it's saved in db) a WooCommerce Admin Note, which WooCommerce Admin will take care of outputting and handling.

### Implementation

This PR implements some conditional logic to achieve this, therefore existing plugins calling `SV_WC_Admin_Notice_Handler::add_admin_notice()` will already support admin notes if WooCommerce Admin is present.

The handler takes care of normalizing some arguments that are expected by admin notes, which differ from the arguments normally accepted by `add_admin_notice()`.  `add_admin_notice()` is made able to handle both legacy and new arguments. Our plugins can use either set of arguments and the handler will convert them for appropriate use according to WooCommerce Admin presence. 

Newer usages of `add_admin_notice()` are encouraged to use the new set of arguments, as the admin notes allow for additional fine tuning. The accepted arguments are the same as the `WC_Admin_Note` object constructor, which I'm reproducing below. Whenever I added an inline comment, there is a default value the handler will take care of; when the comment is not present, the default value is handled by WooCommerce Admin:

```php
// by calling this existing method we can already set an admin note, with or without these new additional arguments
SV_WC_Admin_Notice_Handler::add_admin_notice( $message_id, $message_content, [
	'name'          => '-', // slug, will default to $message_id
	'type'          => self::E_WC_ADMIN_NOTE_INFORMATIONAL, // milestones will be info, other notes will try to match the current notice_class attribute
	'locale'        => 'en_US', // we should have WC admin deal with it for now
	'title'         => '-', // will default to plugin name
	'content'       => '-', // will default on $message_content
	'icon'          => 'info', // will default to matching 'type'
	'content_data'  => new stdClass(),
	'status'        => self::E_WC_ADMIN_NOTE_UNACTIONED, // this means the note wasn't dismissed, by default
	'source'        => 'woocommerce', // this will match the plugin name, e.g. 'woocommerce-memberships'
	'date_created'  => '0000-00-00 00:00:00', // this will default to current time
	'date_reminder' => '', // this is handled by WC Admin in case the user snoozes the note
	'is_snoozable'  => false, // defaults to false if the note can be dismissed, true otherwise
	'actions'       => [], // if this key is not passed at all, and the note is dismissible, a "Dismiss" action will be set
] );
```

This is all that should be needed to start using notes. The admin notice handler class in this PR also comes new public methods (also used internally) to do basic CRUD operations with notes exclusively.

Existing methods have been updated to detect whether to look for notes or notices and return result consistent with existing return types, as well accepted method args.

### Types

Note types replace the `notice_class` which our admin notices have been using to color code the type of notice (warning, error, etc.).

Also Admin Notes use color coding, however, the `type` parameter will also determine where the note will appear. See following table:

|Type   |Suggested usage                     |Position   |Color
|-------|------------------------------------|-----------|-----
|info   |General info, milestones            |Inbox      |gray
|update |Plugin updates, important info      |Foreground |blue
|warning|General notice, non critical errors |Inbox      |orange/yellow
|error  |Errors demanding attention          |Foreground |red

Admin notes can appear in a foreground position, close to where WordPress notices use to show. Except they're stacked and a merchant has to cycle through them (most recent on top). Alternatively, minor messages (info, notice) may appear in an "inbox" sidebar in background, which the merchant can access to by clicking. Notes appearing  in inbox will make use of the `icon` attribute (see [Gridicons](https://github.com/Automattic/gridicons)).

### Status

Admin notes can have a status. Generally supported statuses are three: `unactioned`, `actioned`, `snoozed`. Most notes will start as `unactioned`: the user (or the plugin) has not taken any action (or the plugin hasn't dismissed the note by logic). Some notes can also be snoozed if a `is_snoozable` parameter is `true`. This generally applies to notes that appear in foreground: `snoozable` notes can be snoozed for different lengths of time. 

### Actions

Admin notes can display buttons triggering actions. Normally these actions are simply URLs that the button will follow (but nothing forbids to attach some advanced behavior here). A note can have any number of actions, or none. If none, the note needs to be dismissed via PHP code or REST API. 

### Snoozing a note

If `is_snoozable` was set to `true` an admin note shown in foreground can be snoozed by the merchant. WooCommerce Admin takes care of that (unsnoozing is done via Action Scheduler). If a note is snoozed it is temporarily dismissed from view and its status is moved to `snoozed`.

We should probably use this for nags that cannot (or shouldn't) be dismissed via action button. For example, deprecation notices. 

### Dismissing and deleting notes

Actioned notes (notes where the user has clicked an action button which triggers an `actioned` status in the note) are dismissed from view, however are retained in database. WooCommerce Admin does not look like having an automated mechanism for clearing old dismissed notes by this time and it's up to a plugin to delete its notes from db.

There is an action `woocommerce_admin_note_action` or more specific `woocommerce_admin_note_action_{$action_name}` that could be useful to hook to to automatically delete notes that have been "actioned" by an action button. It is not clear what is the recommended way to delete these notes permanently to avoid clutter in db. There is an [open issue](https://github.com/woocommerce/woocommerce-admin/issues/2505) on the WooCommerce Admin repository to introduce a proper dismissal default action, so this seems to be an open topic.

Finally, actions attached to a note is treated as meta data which gets its own table. Even if actions share the same name, new actions are created for the same note, even if they share the name of actions of other notes. When a note is deleted, its associated actions are deleted from db as well.

### Content data

Notes can also carry objects in one of their properties. This is additional, optional, data that could be used for further manipulation when an action is taken, or to carry data that is not a string and cannot be output in the content body. This type of content could be very useful for support purposes, among other usages, because, for example, the data attached to the note triggered by an error that prompted the note creation could be sent as part of a support request and processed by our support bot.

### Display

Notes are displayed anywhere WooCommerce Admin can output its views. Normally, these are all core WooCommerce admin pages (e.g. products, orders, settings...) and the newly introduced Analytics pages.

Third parties and extensions need to declare support to WooCommerce Admin to have content displayed on them, using the `wc_admin_connect_page()` function and maybe further adjusting the `wc_admin_connect_page_options` filter. The passed argument for each should contain the following data:

```php
/**
 * Connects a page to WooCommerce Admin.
 *
 * Use the `js_page` option to determine if registering.
 *
 * @param array $options {
 *   Array describing the page.
 *
 *   @type string       id           Id to reference the page.
 *   @type string|array title        Page title. Used in menus and breadcrumbs.
 *   @type string|null  parent       Parent ID. Null for new top level page.
 *   @type string       screen_id    The screen ID that represents the connected page. (Not required for registering).
 *   @type string       path         Path for this page. E.g. admin.php?page=wc-settings&tab=checkout
 *   @type string       capability   Capability needed to access the page.
 *   @type string       icon         Icon. Dashicons helper class, base64-encoded SVG, or 'none'.
 *   @type int          position     Menu item position.
 *   @type boolean      js_page      If this is a JS-powered page.
 * }
 */
wc_admin_connect_page( $options );
```

## Caveats

Some things to keep in mind when dealing with notes vs notices:

* Notices can be dismissed per user; notes are shown to all shop managers and their status changes for all
* Notices can be shown non any screen with minimal tweaking; notes will only be shown on screens that have been connected to WC Admin. Once connected it's harder not to show them on certain screens but keep them in others.
* WooCommerce Admin is still receiving a lot of updates and therefore it may be a bit premature to add support to WooCommerce Admin Notes into Framework before they are even merged into a WooCommerce beta.

### Connect pages to WooCommerce Admin

To fully support admin notes, our extensions need also to support to WooCommerce admin, otherwise its screens/cards/menu bar won't show on non-connected screens. 

To do so  one can use `wc_admin_connect_page()`. To detect whether a page has been connected, there's the function `wc_admin_is_connected_page()`. The problem with the latter is that within its logic, WooCommerce Admin makes use of `get_current_screen()` which is not always available when our plugins/framework need to check whether the admin notes system is available. 

So we would have to revise/check when our plugins add their notices, to be conditionally turned in notes, and see if this function is available and the timing to connect to WooCommerce Admin is right. Or possibly, we could add more Framework methods, connected our pages very early and then check internally if those are going to be connected, so our `SV_WC_Admin_Notice_Handler::should_use_admin_notes()` returns true for them.

### Messages handler

This PR doesn't deal yet with WordPress messages, but notes are both notices and messages. To display the latter in WooCommerce Admin pages, they also need to be adapted into notes.

## QA

- [ ] To test an implementation, a plugin needs to point its composer file to use this dev branch of the Framework rather than a release branch
- [ ] Try having such plugin setting admin notices: they should be displayed as admin  notes

### DIFF
--- a/woocommerce/Lifecycle.php
+++ b/woocommerce/Lifecycle.php
@@ -87,7 +87,7 @@ class Lifecycle {
 		}
 
 		// catch any milestones triggered by action
-		add_action( 'wc_' . $this->get_plugin()->get_id() . '_milestone_reached', array( $this, 'trigger_milestone' ), 10, 4 );
+		add_action( 'wc_' . $this->get_plugin()->get_id() . '_milestone_reached', [ $this, 'trigger_milestone' ], 10, 4 );
 	}
 
 
@@ -441,6 +441,7 @@ class Lifecycle {
 				[
 					'name'    => 'dismiss',
 					'label'   => __( 'Dismiss', 'woocommerce-plugin-framework' ),
+					'url'     => '#',
 				]
 			];
 		}
--- a/woocommerce/Lifecycle.php
+++ b/woocommerce/Lifecycle.php
@@ -313,9 +313,11 @@ class Lifecycle {
 
 			if ( $message ) {
 
-				$this->get_plugin()->get_admin_notice_handler()->add_admin_notice( $message, $id, array(
+				$this->get_plugin()->get_admin_notice_handler()->add_admin_notice( $message, $id, [
 					'always_show_on_settings' => false,
-				) );
+					'is_snoozable'            => false,
+					'type'                    => 'info',
+				] );
 
 				// only display one notice at a time
 				break;
--- a/woocommerce/Lifecycle.php
+++ b/woocommerce/Lifecycle.php
@@ -87,7 +87,7 @@ class Lifecycle {
 		}
 
 		// catch any milestones triggered by action
-		add_action( 'wc_' . $this->get_plugin()->get_id() . '_milestone_reached', array( $this, 'trigger_milestone' ), 10, 3 );
+		add_action( 'wc_' . $this->get_plugin()->get_id() . '_milestone_reached', array( $this, 'trigger_milestone' ), 10, 4 );
 	}
 
 
@@ -301,6 +301,13 @@ class Lifecycle {
 				continue;
 			}
 
+			$args = [];
+
+			if ( is_array( $message ) ) {
+				$message = key( $message );
+				$args    = current( $message );
+			}
+
 			/**
 			 * Filters a milestone notice message.
 			 *
@@ -311,13 +318,19 @@ class Lifecycle {
 			 */
 			$message = apply_filters( 'wc_' . $this->get_plugin()->get_id() . '_milestone_message', $this->generate_milestone_notice_message( $message ), $id );
 
-			if ( $message ) {
+			if ( '' !== $message && is_string( $message ) ) {
 
-				$this->get_plugin()->get_admin_notice_handler()->add_admin_notice( $message, $id, [
-					'always_show_on_settings' => false,
-					'is_snoozable'            => false,
-					'type'                    => 'info',
-				] );
+				/**
+				 * Filters milestone notice arguments.
+				 *
+				 * @since 5.4.1-dev.1
+				 *
+				 * @param array $args optional arguments
+				 * @param string $id milestone ID
+				 */
+				$args = (array) apply_filters( 'wc_' . $this->get_plugin()->get_id() . '_milestone_args', $this->generate_milestone_notice_args( $args ), $id );
+
+				$this->get_plugin()->get_admin_notice_handler()->add_admin_notice( $message, $id, $args );
 
 				// only display one notice at a time
 				break;
@@ -344,16 +357,17 @@ class Lifecycle {
 	 * @param string $id milestone ID
 	 * @param string $message message to display to the user
 	 * @param string $since the version since this milestone has existed in the plugin
+	 * @param array $args optional arguments used in admin notes
 	 * @return bool
 	 */
-	public function trigger_milestone( $id, $message, $since = '1.0.0' ) {
+	public function trigger_milestone( $id, $message, $since = '1.0.0', array $args = [] ) {
 
 		// if the plugin was had milestones before this milestone was added, don't trigger it
 		if ( version_compare( $this->get_milestone_version(), $since, '>' ) ) {
 			return false;
 		}
 
-		return $this->register_milestone_message( $id, $message );
+		return $this->register_milestone_message( $id, $message, $args );
 	}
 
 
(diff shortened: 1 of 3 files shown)


unfulvio:
hey @ChaseWiseman in the [message above](https://github.com/skyverge/wc-plugin-framework/pull/339#issue-300663364) you can see a writeup for my current take at WooCommerce Admin Notes. This PR offers a drop in integration that allows using the existing `add_admin_notice()` to set an admin note instead, and nothing else is really needed. At the same time, I also think that we should hold off a bit before considering adopting notes in Framework as there are still changes going on. Nonetheless, you could take a look at this PR by connecting it to any of our plugins targeting this dev branch (they need to update to `5_4_1` namespaces). Thank you!

ChaseWiseman:
Closing as for now we decided not to move forward with the drop-in replacement.

PR #341, opened 2019-08-15
[ch14390] Replacement screen helper methods

unfulvio:
## Summary

We had from time to time cases where `get_current_screen()` results in a PHP error for function not existing, since this function is not always made available in WordPress, even in admin context. 

The best practice would be using the global `$current_screen` and then checking if it's not null and access the `id` property of the `WP_Screen` class. 

So, we can have a couple of helper methods to handle these common checks for us.

### Details

Adds:

* `SV_WC_Helper::get_current_screen()` replacement for `get_current_screen()` which will never cause a function not existing PHP error

* `SV_WC_Helper::is_current_screen( $id )` which will check if the current screen has a specified ID (can work with other properties to via secondary optional argument) - this simplifies bunch of common ops that require to declare `global $current_screen` check if is not null/has `id` prop and then compare with another variable.

### Resources

A bunch of cases we had concerning current screen function, missing checking if function exists, or if global is not null (but I'm sure there are/were more):

* https://app.clubhouse.io/skyverge/story/11754/fatal-error-when-using-get-current-screen
* https://app.clubhouse.io/skyverge/story/14280/calling-get-current-screen-causes-error-emails-to-be-sent
* https://app.clubhouse.io/skyverge/story/8674/throws-internal-server-error-with-woocommerce-3-6

### DIFF
--- a/woocommerce/admin/abstract-sv-wc-plugin-admin-setup-wizard.php
+++ b/woocommerce/admin/abstract-sv-wc-plugin-admin-setup-wizard.php
@@ -145,9 +145,7 @@ abstract class Setup_Wizard {
 	 */
 	public function add_admin_notices() {
 
-		$current_screen = get_current_screen();
-
-		if ( ( $current_screen && 'plugins' === $current_screen->id ) || $this->get_plugin()->is_plugin_settings() ) {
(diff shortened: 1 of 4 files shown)


jdeeburke:
Once this is merged, update [this Social Login PR](https://github.com/skyverge/wc-plugins/pull/3179) to update the FW and use this new fix

PR #342, opened 2019-08-15
Protect against notices when get_current_screen() returns null

jdeeburke:
### [CH 14390](https://app.clubhouse.io/skyverge/story/14390)

# Summary

`get_current_screen()` sometimes returns null, and our two usages of it in the framework can throw notices in those rare instances, one of which is visible in the [Social Login setup wizard](http://cloud.skyver.ge/a895da336fe6).

even though `get_current_screen()` checks the global variable `$current_screen`, I chose to check the same global rather than continue to use `get_current_screen()` because it's much simpler and more concise to just call `isset( $current_screen->id )` rather than `is_callable( 'get_current_screen' ) && isset( get_current_screen()->id )`

### DIFF
--- a/woocommerce/admin/abstract-sv-wc-plugin-admin-setup-wizard.php
+++ b/woocommerce/admin/abstract-sv-wc-plugin-admin-setup-wizard.php
@@ -144,10 +144,9 @@ abstract class Setup_Wizard {
 	 * @since 5.2.2
 	 */
 	public function add_admin_notices() {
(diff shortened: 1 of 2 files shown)


unfulvio:
how about #341 ?

jdeeburke:
@unfulvio Hah! Looks like we were thinking the same thing, we posted our PRs 1 minute apart.

Yours is a better solution, I'm gonna close this one :)

PR #344, opened 2019-08-23
[ch10558] Alert customer when key configuration is missing

jdeeburke:
# Summary

Adds an admin notice when a gateway is enabled, but not configured properly and can't accept payments.

## QA

- [ ] Enable a gateway without configuring it - see the notice
- [ ] Configure the gateway and save the settings - notice is gone
- [ ] Override `$gateway->get_not_configured_error_message()` for a gateway and see that the notice is changed to that overridden message

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -4152,6 +4152,26 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 	}
 
 
+	/**
+	 * Returns the error message for display if the gateway is not configured.
+	 *
+	 * @since 5.4.2-dev.1
+	 *
+	 * @return string
+	 */
+	public function get_not_configured_error_message() {
+
+		return sprintf(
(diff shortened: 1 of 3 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -4152,6 +4152,26 @@ public function get_checkout_order_received_order_id() {
...
+	 *
+	 * @since 5.4.2-dev.1
+	 *
+	 * @return string
+	 */
+	public function get_not_configured_error_message() {
+
+		return sprintf(
+			/* translators: %1$s - gateway name, %2$s - <a> tag, %3$s - </a> tag, %4$s - <a> tag */
+			__( 'Heads up! %1$s is not fully configured and cannot accept payments. Please %2$sreview the documentation%3$s and configure the %4$sgateway settings%3$s.', 'woocommerce-plugin-framework' ),
```
unfulvio: I know this is technically correct, however since translations may be handled by non-devs I believe that using the same placeholder multiple times can be error prone - %3$s in this case, containing the closing </a> tag. In other occasions I see we just introduce another placeholder, so each pair of opening/closing tags have their own </a> represented by a different placeholder.

PR #346, opened 2019-09-03
[ch15938] Add country helper class

jdeeburke:
# Add helper class for dealing with country codes

**CH Story: [CH15938](https://app.clubhouse.io/skyverge/story/15938/add-country-helper-class)**

## Details

There is one country code utility function in the framework already -- [`SV_WC_Helper::convert_country_code()`](https://github.com/skyverge/wc-plugin-framework/blob/c458685ac4a7a50cb591b830698d01caf4d3730f/woocommerce/class-sv-wc-helper.php#L1005) and [two](https://github.com/skyverge/wc-plugins/blob/86b586f25ef99ddd9af3ca6eef81d3325808f63e/woocommerce-gateway-realex-redirect/includes/hpp/Payment_Form_Helper.php#L159) that are needed for both of the Global Payments plugins. This PR extracts these helpers into their own static class along with a handful of utility functions for converting back and forth between each ISO 3166 type.

The existing helper function has been deprecated and now calls the new helper.

### DIFF
--- a/woocommerce/class-sv-wc-country-helper.php
+++ b/woocommerce/class-sv-wc-country-helper.php
@@ -1,0 +1,658 @@
+<?php
+/**
+ * WooCommerce Plugin Framework
+ *
+ * This source file is subject to the GNU General Public License v3.0
+ * that is bundled with this package in the file license.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://www.gnu.org/licenses/gpl-3.0.html
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@skyverge.com so we can send you a copy immediately.
+ *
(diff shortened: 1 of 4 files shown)


woocommerce/class-sv-wc-country-helper.php
```diff
@@ -0,0 +1,658 @@
...
+
+	/**
+	 * SkyVerge Country Helper Class
+	 *
+	 * The purpose of this class is to centralize country-related utility
+	 * functions that are commonly used in SkyVerge plugins
+	 *
+	 * @since 5.4.3-dev.1
+	 */
+	class SV_WC_Country_Helper {
```
ChaseWiseman: For these new classes, I think we can ditch the `SV_WC_` prefix and also name the file as `Country_Helper.php`

PR #348, opened 2019-09-17
v5.5.0

unfulvio:
# Summary

Drops WC < 3.0 and PHP < 5.6 backwards compatibility methods.

### Stories

* [ch17418](https://app.clubhouse.io/skyverge/story/17418/remove-wc-2-6-compatibility-code) Remove WC 2.6 Compatibility code
* [ch17414](https://app.clubhouse.io/skyverge/story/17414/add-global-wrappers-for-common-helper-methods) Add global wrappers for common helper methods ([no implementation](https://app.clubhouse.io/skyverge/story/17414/add-global-wrappers-for-common-helper-methods), see [wiki](https://github.com/skyverge/wc-plugin-framework/wiki/Class-aliases-and-template-functions) instead)
* [ch14044](https://app.clubhouse.io/skyverge/story/14044/add-helper-method-for-wc-get-template-that-passes-default-path-automatically) Adds `wc_get_template()` helper method in main plugin class

#### Global wrappers

We figured we could use class aliases defined such as `class_alias( SV_WC_Helper::class, 'WC_Plugin_Name_Helpers' );` inside each plugin that needs to use helper/compatibility functions inside template files or other places where namespaces aren't safe to use. Instead of defining these aliases inside the framework, they can be defined case by case by plugins that need them. See [ClubHouse discussion](https://app.clubhouse.io/skyverge/story/17414/add-global-wrappers-for-common-helper-methods#activity-17564).

#### Add `SV_WC_Plugin::load_template()`

Adds a `SV_WC_Plugin::load_template()` helper method that is a replacement for `wc_get_template()`. The method accepts a template part/name (string) and an optional array of arguments to pass to the template. Plugins can replace `wc_get_template()` with this method instead.

The main difference is that we pass automatically a default path argument to `wc_get_template()` which was requested by [Leho on Teams](https://github.com/skyverge/wc-memberships-addons/issues/67), plus in this way we can ensure consistency.

#### Deprecations

No methods have been removed in this version, however, there are several backwards compatibility methods meant for backporting WC 3.0 CRUD methods to WC 2.6 and below. Since we no longer support 2.6, these can be marked as `@deprecated` and a `wc_deprecated_function()` notice will be issued. By version 6.0 we should probably permanently remove deprecated methods. We can keep the `\WC_Data` handlers probably as it's possible that in future versions WC core will add some methods that we wish to backport (currently there's one for the order edit URL).

Plugins wishing to update to this version of the Framework will need to remove any usage of deprecated methods to avoid triggering notices.

### List of deprecated methods

See newly created [Deprecated methods](https://github.com/skyverge/wc-plugin-framework/wiki/Deprecated-methods) page on wiki.

### DIFF
--- a/woocommerce-framework-plugin-loader-sample.php
+++ b/woocommerce-framework-plugin-loader-sample.php
@@ -58,7 +58,7 @@ class SV_WC_Framework_Plugin_Loader {
 	const MINIMUM_WC_VERSION = '3.0.9';
 
 	/** SkyVerge plugin framework version used by this plugin */
-	const FRAMEWORK_VERSION = '5.4.3'; // TODO: framework version
+	const FRAMEWORK_VERSION = '5.5.0'; // TODO: framework version
 
 	/** the plugin name, for displaying notices */
 	const PLUGIN_NAME = 'WooCommerce Framework Plugin'; // TODO: plugin name
@@ -157,7 +157,7 @@ class SV_WC_Framework_Plugin_Loader {
 	 *
 	 * @since 1.0.0
 	 */
-	protected function load_framework() {
+	private function load_framework() {
 
 		if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\' . $this->get_framework_version_namespace() . '\\SV_WC_Plugin' ) ) {
 			require_once( plugin_dir_path( __FILE__ ) . 'lib/skyverge/woocommerce/class-sv-wc-plugin.php' );
@@ -177,7 +177,7 @@ class SV_WC_Framework_Plugin_Loader {
 	 *
 	 * @return string
 	 */
-	protected function get_framework_version_namespace() {
+	private function get_framework_version_namespace() {
 
 		return 'v' . str_replace( '.', '_', $this->get_framework_version() );
 	}
@@ -190,7 +190,7 @@ class SV_WC_Framework_Plugin_Loader {
 	 *
 	 * @return string
 	 */
-	protected function get_framework_version() {
+	private function get_framework_version() {
 
 		return self::FRAMEWORK_VERSION;
 	}
@@ -200,6 +200,8 @@ class SV_WC_Framework_Plugin_Loader {
 	 * Checks the server environment and other factors and deactivates plugins as necessary.
 	 *
 	 * Based on http://wptavern.com/how-to-prevent-wordpress-plugins-from-activating-on-sites-with-incompatible-hosting-environments
+	 *
+	 * @internal
 	 *
 	 * @since 1.0.0
 	 */
@@ -213,8 +215,11 @@ class SV_WC_Framework_Plugin_Loader {
 		}
 	}
 
+
 	/**
 	 * Checks the environment on loading WordPress, just in case the environment changes after activation.
+	 *
+	 * @internal
 	 *
 	 * @since 1.0.0
 	 */
@@ -231,6 +236,8 @@ class SV_WC_Framework_Plugin_Loader {
 
 	/**
 	 * Adds notices for out-of-date WordPress and/or WooCommerce versions.
+	 *
+	 * @internal
 	 *
 	 * @since 1.0.0
 	 */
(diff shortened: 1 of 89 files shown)


woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -517,33 +468,25 @@ public static function has_shipping_address( \WC_Order $order ) {
...
 	 * @return array $item_meta {
 	 *     @type string $label meta field label
-	 *     @type mixed  $value meta value
+	 *     @type mixed $value meta value
  	 * }
 	 */
-	public static function get_item_formatted_meta_data( $item, $hideprefix = '_', $include_all = false ) {
+	public static function get_item_formatted_meta_data( $item, $hide_prefix = '_', $include_all = false ) {
 
-		if ( SV_WC_Plugin_Compatibility::is_wc_version_gte_3_1() && $item instanceof \WC_Order_Item ) {
```
ChaseWiseman: we'd still need to keep this check, as we don't yet require 3.1

woocommerce/compatibility/abstract-sv-wc-data-compatibility.php
```diff
@@ -22,56 +22,43 @@
...
-			if ( is_callable( array( $object, "get_{$prop}" ) ) && 'view' === $context ) {
-				$value = $object->{"get_{$prop}"}();
-			} else {
-				$value = $object->$prop;
-			}
-		}
+		wc_deprecated_function( __METHOD__, '5.5.0', '\WC_Data methods' );
 
-		return $value;
+		return $object->{"get_{$prop}"}( $context );
```
ChaseWiseman: should probably keep the `is_callable()` check

woocommerce-framework-plugin-loader-sample.php
```diff
@@ -177,7 +177,7 @@ protected function load_framework() {
 	 *
 	 * @return string
 	 */
-	protected function get_framework_version_namespace() {
+	private function get_framework_version_namespace() {
```
ChaseWiseman: come to think of it, how about making this method public? Seems like we could provide it in case someone needs to get the namespace dynamically

woocommerce-framework-plugin-loader-sample.php
```diff
@@ -190,7 +190,7 @@ protected function get_framework_version_namespace() {
 	 *
 	 * @return string
 	 */
-	protected function get_framework_version() {
+	private function get_framework_version() {
```
ChaseWiseman: same here, public?

unfulvio:
@ChaseWiseman I've added a comprehensive list of deprecated methods in this PR initial message. If you're ok with the list, and once this is merged, the list could be transferred to wiki. The list includes deprecated methods from previous versions. 

Additionally, since this PR will prompt plugins to make some internal changes, how about we introduce some replacement/alias for `SV_WC_Helper` methods `get_post()` and `get_request()`. We discussed this before, especially `get_post()` has a confusing name with WP own `get_post()` and `get_request()` is often used in API requests contexts, but here it can be from a HTTP form as well. I was thinking `get_posted_data()` or perhaps `get_http_post_data()` and `get_http_request_data()`. Thoughts?


There are some methods in `SV_WC_Helper` like `get_order_line_items()` and `is_order_virtual()` that seem to pertain more to the `SV_WC_Order_Compatibility` class. How about we limit the concern for `SV_WC_Helper` to strictly generic PHP helpers?

ChaseWiseman:
@unfulvio list looks good! thanks for that. yeah let's add it to a wiki page on this repo

>Additionally, since this PR will prompt plugins to make some internal changes, how about we introduce some replacement/alias for SV_WC_Helper methods get_post() and get_request(). 

Fine with me. `get_posted_data()` looks good imo

>There are some methods in SV_WC_Helper like get_order_line_items() and is_order_virtual() that seem to pertain more to the SV_WC_Order_Compatibility class. How about we limit the concern for SV_WC_Helper to strictly generic PHP helpers?

Hmmm, I don't think they belong in the compatibility class tbh. Those methods aren't wrappers for compatibility - they're utility helper methods. If there were a ton of them then we could add a `Order_Helper` class but I don't think that's warranted right now.

PR #349, opened 2019-09-17
[ch16778] [ch16779] Debug API endpoints

unfulvio:
## Summary

Implements 2 API endpoints for remote debugging of WC plugins:

* `wc/v3/<plugin_id>/debug` route to get the current debug mode of the plugin (and any of its gateways), or update the debug mode
* `wc/v3/<plugin_id>/log` route to list logs for the plugin (and any of its gateways)

### Related stories

* [16778](https://app.clubhouse.io/skyverge/story/16778/add-api-endpoint-to-toggle-debug-logging)
* [16779](https://app.clubhouse.io/skyverge/story/16779/add-an-api-endpoint-to-retrieve-debug-logs)

## Details

The PR extends the current `REST_API` handler to initialize the two new endpoints in all plugins that will use the framework at this version. Currently only the `v3` namespace of the WooCommerce REST API is supported, but the code is structured in a way that is possible to extend support to newer versions of the REST API in the future. 

### Debug mode

The debug endpoint will list the current debug mode used by the plugin and, if a gateway, of any gateways added by the plugin.

#### Plugins

Not all plugins have a debug mode, but we do make use of log files for a parent plugin, for example to record data during upgrade routines, or other notable errors. As such, child plugins that do not define a debug mode or offer a way to toggle it, will be assumed as if they had a debug mode set to log. In this case, plugins implementing this do not need to do anything. 

Plugins that may have a debug mode, can implement `get_debug_mode()` and `set_debug_mode()` methods to get and set the corresponding mode. 

#### Gateways

Gateways do not need to do anything either as they have a more structured debug mode and known debug modes. The parent plugin that handles the individual gateways, should it have a special debug mode unrelated to gateways, can implement the two methods mentioned before just like any other plugin.

When setting a debug mode from a POST/PUT/PATCH request, the data format to be sent matches the one from GET responses. 

### Logs

The log endpoint will list all the logs found for the matching plugin, and its gateways (if any). This is because, symmetrically to debug mode, some plugins can have a separate parent plugin log file and log files specifically for gateways. 

The response will normalize the output regardless if the installation is using log files or log entries in database using `WC_Log_Handler_DB`.

Plugins do not need to do anything to support this endpoint. 

### Debug modes normalization

While gateways have structured debug modes defined in constants, plugins do not have these. For example, MailChimp uses only "yes" or "no" as debug mode. Other plugins may use "enabled" or "disabled" and so on. 

- Should we normalize these? Most likely they are going to be "log" or "off" statuses, but we can't predict if some plugins will require a different mode. 

- Otherwise, should we list the accepted debug modes in a separate property of the response object?

### Log files size

Could there be any issue to send large file contents via REST API? WC should limit these by date/size but they can still grow large.

_In theory_ one could attempt to parse the debug log contents, since the logs always begin with a date. However, this could be unreliable since it's not markup/structured data and data could be missing in the response.

## Resources

The current PR is implemented in 2 plugins ready for testing:

- [MailChimp Sync](https://github.com/skyverge/wc-memberships-addons/pull/112/files)
- [Authorize.Net](https://github.com/skyverge/wc-plugins/pull/3223)

### DIFF
--- a/woocommerce/rest-api/Debug_Controller.php
+++ b/woocommerce/rest-api/Debug_Controller.php
@@ -24,14 +24,13 @@
 
 namespace SkyVerge\WooCommerce\PluginFramework\v5_5_0\REST_API;
 
-use SkyVerge\WooCommerce\PluginFramework\v5_5_0\SV_WC_Payment_Gateway;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_0\SV_WC_Payment_Gateway_Plugin;
 use SkyVerge\WooCommerce\PluginFramework\v5_5_0\SV_WC_Plugin;
-use SkyVerge\WooCommerce\PluginFramework\v5_5_0\SV_WC_Plugin_Exception;
 
 defined( 'ABSPATH' ) or exit;
 
 if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_5_0\\REST_API\\Debug_Controller' ) ) :
+
 
 /**
  * The plugin REST API Debug endpoint.
@@ -42,7 +41,7 @@ abstract class Debug_Controller extends \WC_REST_Controller {
 
 
 	/** @var SV_WC_Plugin main instance */
-	private $plugin;
+	protected $plugin;
 
 	/** @var string endpoint namespace */
 	protected $namespace;
@@ -84,13 +83,13 @@ abstract class Debug_Controller extends \WC_REST_Controller {
 				'permission_callback' => [ $this, 'get_item_permissions_check' ],
 			],
 			// UPDATE (toggle) the debug mode
-			array(
+			[
 				'methods'             => \WP_REST_Server::EDITABLE,
 				/** @see \WC_REST_Controller::update_item() */
 				'callback'            => [ $this, 'update_item' ],
 				/** @see \WC_REST_Controller::update_item_permissions_check() */
 				'permission_callback' => [ $this, 'update_item_permissions_check' ],
-			),
+			],
 		], true );
 	}
 
@@ -129,14 +128,18 @@ abstract class Debug_Controller extends \WC_REST_Controller {
 		];
 
 		if ( $plugin instanceof SV_WC_Payment_Gateway_Plugin ) {
+
+			$debug_mode['gateways'] = [];
+
 			foreach ( $plugin->get_gateways() as $gateway ) {
-				$debug_mode[ $gateway->get_id() ] = $gateway->get_debug_mode();
+
+				$debug_mode['gateways'][ $gateway->get_id() ] = $gateway->get_debug_mode();
 			}
 		}
 
 		return rest_ensure_response( [
 			'debug_mode' => $debug_mode,
-			'debug_log'  => defined( 'WC_LOG_HANDLER' ) ? WC_LOG_HANDLER: 'file',
+			'debug_log'  => defined( 'WC_LOG_HANDLER' ) ? WC_LOG_HANDLER : 'file',
 		] );
 	}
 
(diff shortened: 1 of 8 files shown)


woocommerce/rest-api/Debug_Controller.php
```diff
@@ -0,0 +1,256 @@
...
+	 *
+	 * @since 5.5.0-dev
+	 *
+	 * @param SV_WC_Plugin $plugin main instance
+	 */
+	public function __construct( SV_WC_Plugin $plugin ) {
+
+		$this->plugin    = $plugin;
+		$this->namespace = 'wc/v1';
+		$this->rest_base = $plugin->get_id();
```
ChaseWiseman: let's prefix this with `wc_` for consistency with how the plugin IDs are used throughout the framework

woocommerce/rest-api/Debug_Controller.php
```diff
@@ -0,0 +1,256 @@
...
+	 *
+	 * @param \WP_REST_Request $request API request
+	 * @return \WP_Error|\WP_REST_Response API response
+	 */
+	public function get_item( $request ) {
+
+		$plugin     = $this->get_plugin();
+		$plugin_id  = $plugin->get_id();
+		$debug_mode = [
+			'plugin' => [
```
ChaseWiseman: any need to wrap this in `plugin`? We're already requesting a specific plugin via the route, so we know the response will contain the plugin's data. `gateways` can still be the sub object, but I don't think we'd need to separate this out.

woocommerce/rest-api/class-sv-wc-plugin-rest-api.php
```diff
@@ -24,9 +24,12 @@
 
 namespace SkyVerge\WooCommerce\PluginFramework\v5_5_0;
 
+use SkyVerge\WooCommerce\PluginFramework\v5_5_0\REST_API\Debug_Controller;
+use SkyVerge\WooCommerce\PluginFramework\v5_5_0\REST_API\Log_Controller;
+
 defined( 'ABSPATH' ) or exit;
 
-if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_5_0\\REST_API' ) ) :
```
ChaseWiseman: looks like an accidental change?

woocommerce/rest-api/class-sv-wc-plugin-rest-api.php
```diff
@@ -134,12 +154,87 @@ protected function get_system_status_data() {
...
+
+	/**
+	 * Gets the debug controller endpoint handler instance.
+	 *
+	 * @since 5.5.0-dev
+	 *
+	 * @param string $api_version API version for the controller to be returned
+	 * @return null|Debug_Controller
+	 */
+	public function get_debug_controller_instance( $api_version = 'v3' ) {
```
ChaseWiseman: one thing to be mindful of is that we need to give plugins the ability to override the controller class easily, in case they need to use their own child class instead.

Perhaps we could use an additional method to get the controller class name, which would be overridable without having to duplicate the logic inside this method?

woocommerce/rest-api/Debug_Controller.php
unfulvio: but that's not the case in memberships though, this is not a PHP/framework context, rather API context - the endpoint would look like `wc/v3/wc_...` seems a bit weird to me

woocommerce/rest-api/Debug_Controller.php
unfulvio: My rationale was that there could be debug modes and logs for different entities, and to keep everything in key-value paris for each debug component for consistency. 

What if a non-gateway plugin has a debug mode for itself plus a third part API? Seems more consistent to me, to expect that the debug data for the plugin specifically is in the key `plugin`; then `gateways` and so on. 

This makes it also more explicit when actors need to call each identity, you can see what debug mode they are touching. Being a sensitive area, I wanted the minimum ambiguity.  

Perhaps one could add a new key (not sure of the naming -- `components`?) that lists the available debug components... Like `plugin`, `gateways`... so a third party knows these keys exist?

ChaseWiseman: @unfulvio thanks for more details! I understand what you mean about the different component considerations.

I think breaking things out further is overly complicated for what the original story is trying to achieve - allow third-party apps to toggle debug logging.

I still find the response structure and data to be a bit confusing as-is. Taking a plugin like Memberships as an example (to get away from gateways) - the plugin has no debug mode. As such, I wouldn't expect the response from this endpoint to show the debug mode as `log`, since that isn't really true. The base plugin `->log()` method is purely internal, and doesn't represent any sort of debug state. Now, we could add a debug mode to Memberships in the future, and it wouldn't be a bad idea, but right now there is no such concept.

Similarly, a gateway plugin like Authorize.Net doesn't have a debug more. Its gateways do, but the plugin itself does not.

With that in mind, it seems like we should return something that more accurately represents the plugin's lack of debug mode, like `null`. Also, is there any reason to include the plugin ID again inside `plugin`, when we're already requesting that plugin ID? It seems like that would make sense if we were listing a bunch of plugins and their debug modes, but this is data for a specific plugin.

Considering a plugin like AvaTax - that plugin _does_ have a debug mode since the API debugging is no separated from the plugin itself.

unfulvio: hey @ChaseWiseman fair points, I've updated the response to omit the `plugin` key; additionally, I'm listing the available debug modes in a `debug_modes` key; this matches a PHP method that can also be used to validate modes being set or when listing modes in settings for example - I've updated MailChimp Sync test release with the new version; Authorize.Net didn't require anything, just need to update the FW from composer

ChaseWiseman:
Closing in favor of the larger settings API epic: https://app.clubhouse.io/skyverge/epic/23790/framework-settings-api

PR #356, opened 2019-11-11
Do not reduce stock levels for held orders 

unfulvio:
## Summary

The framework will manually reduce stock for orders which have a held status.

However, WooCommerce 3.5+ will do this already, resulting in twice reduced stock for such orders. 

### Story

- [ch20837](https://app.clubhouse.io/skyverge/story/20837/stock-reduced-twice-when-order-is-held-but-payment-not-completed)

### Details

This PR will introduce a WC 3.5 check before reducing stock levels for held orders.

## QA

- [x] Follow QA steps in this [PR for Authorize.Net](https://github.com/skyverge/wc-plugins/pull/3328) which generated most support tickets
- [x] Similar [PR for Elavon](https://github.com/skyverge/wc-plugins/pull/3330), also present in support tickets

### Resources

 - [Current WC code to reduce stock for held orders](https://github.com/woocommerce/woocommerce/blob/d9f971627401d46589de088457411663f41135a6/includes/wc-stock-functions.php#L108)
- [Commit in WC to reduce stock, tagged for 3.5.0-beta](https://github.com/woocommerce/woocommerce/commit/70c9cff608761fcd48b57f709059e00b1ffeee38)

### DIFF
--- a/woocommerce/changelog.txt
+++ b/woocommerce/changelog.txt
@@ -1,6 +1,7 @@
 *** SkyVerge WooCommerce Plugin Framework Changelog ***
 
 2019.nn.nn - version 5.5.1
+ * Fix - Do not reduce twice the stock level of a product when the order is held and payment is not completed
 
 2019.10.15 - version 5.5.0
  * Feature - Add a plugin helper method to retrieve a template part while consistently passing the default template path to `wc_get_template()`
--- a/woocommerce/payment-gateway/Handlers/Abstract_Payment_Handler.php
+++ b/woocommerce/payment-gateway/Handlers/Abstract_Payment_Handler.php
@@ -311,10 +311,8 @@ abstract class Abstract_Payment_Handler {
 		$this->get_gateway()->add_payment_gateway_transaction_data( $order, $response );
 
 		if ( $order->has_status( $this->get_held_order_status( $order, $response ) ) ) {
-			// reduce stock for held orders, but don't complete payment (WooCommerce versions 3.5 and newer will do this for us)
-			if ( FrameworkBase\SV_WC_Plugin_Compatibility::is_wc_version_lt( '3.5.0' ) ) {
-				wc_reduce_stock_levels( $order );
-			}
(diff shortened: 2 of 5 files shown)


ChaseWiseman:
@unfulvio I was confused as to why this would suddenly become an issue now, given WC 3.5 is over a year old, and I found that it's not because we call `wc_reduce_stock_levels()`. That method already has [checks in place](https://github.com/woocommerce/woocommerce/blob/d9f971627401d46589de088457411663f41135a6/includes/wc-stock-functions.php#L169) to skip double reduction if called by other code.

It's because of our changes in FW v5.5.0 to [move away from the 3.0 compatibility code](https://github.com/skyverge/wc-plugin-framework/commit/7b1e0148f6d2a3071807aed5724f35d4e2de19f3#diff-615a83511c5f152ac5ccb773eb2841a0R315). That [compatibility method](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/compatibility/class-sv-wc-order-compatibility.php#L426-L431) would call `wc_reduce_stock_levels()` using the order _ID_, whereas in 5.5.0 we now call it using the order object. This results in the method not calling `wc_get_order()`, so it doesn't get a fresh order object with the `reduced_stock` meta set earlier on the status change.

In any case, we shouldn't skip calling `wc_reduce_stock_levels()` based on WC version. Note that the [held order status can be filtered](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/class-sv-wc-payment-gateway-direct.php#L401), so it may not be `on-hold`, which would result in stock not being reduced at all.

So we should either:
- Restore our `wc_reduce_stock_levels()` calls to only pass the order ID
- Switch to using `wc_maybe_reduce_stock_levels()` and see if that covers it

unfulvio:
@ChaseWiseman  I've opted to restore passing the ID than changing behavior, since that worked for us. I've tested with the Authorize.Net branch in https://github.com/skyverge/wc-plugins/pull/3328 and that seems to work well.

PR #364, opened 2019-12-12
Add gateway ID setter & getter methods to token object

unfulvio:
# Summary

This PR adds a `get_gateway_id()` and a `set_gateway_id()` public methods to the `SV_WC_Payment_Gateway_Payment_Token` object.

### QA

- [x] When using `set_gateway_id( (string) $gateway_id )` a `gateway_id` should be added to the `SV_WC_Payment_Gateway_Payment_Token::$data` array.
- [x] When using `get_gateway_id()` a `gateway_id` should be returned as it appears in the `SV_WC_Payment_Gateway_Payment_Token::$data` array.


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -136,6 +137,32 @@ public function get_id() {
...
 
 
+	/**
+	 * Gets the gateway ID for the token.
+	 *
+	 * @since 5.6.0-dev.1
+	 *
+	 * @return string
+	 */
+	public function get_gateway_id() {
```
asseti6: Same here, two separate commits.  The slash in the subject is a good indication of that needing to happen

unfulvio:
hey @deivamagalhaes I just added that in https://github.com/skyverge/wc-plugin-framework/pull/364/commits/3862eea397716db77cbf2fa1c43f17833921cc9b -- will do also for the user_id branch

PR #365, opened 2019-12-12
Add user ID setter & getter methods to token object 

unfulvio:
# Summary

This PR adds a `get_user_id()` and a `set_user_id()` public methods to the `SV_WC_Payment_Gateway_Payment_Token` object.

### QA

- [ ] When using `set_user_id( (int) $user_id )` a `user_id` should be added to the `SV_WC_Payment_Gateway_Payment_Token::$data` array.
- [ ] When using `get_user_id()` a `user_id` should be returned as it appears in the `SV_WC_Payment_Gateway_Payment_Token::$data` array.

### DIFF
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -54,6 +54,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 	 * @var array key-value array to map WooCommerce core token props to framework token `$data` keys
 	 */
 	private $props = [
+		'user_id'      => 'user_id',
 		'is_default'   => 'default',
 		'type'         => 'type',
 		'last4'        => 'last_four',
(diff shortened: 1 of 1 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -136,6 +137,32 @@ public function get_id() {
...
 
 
+	/**
+	 * Gets the ID of the user associated with the token.
+	 *
+	 * @since 5.6.0-dev.1
+	 *
+	 * @return int
+	 */
+	public function get_user_id() {
```
asseti6: @unfulvio these should be two separate commits to be atomic the way we have been briefly discussing

unfulvio:
hey @deivamagalhaes sorry that was a typo on my end (probably I copypasted in the QA instructions the equivalent for gateway_id which does take a string) - user IDs in WooCommerce/WordPress are integers so it's fine here - thanks

PR #368, opened 2019-12-12
Revised Codeception tests approach

unfulvio:
## Summary

This PR brings the test suite used in the framework more in line with what we currently have in Memberships or Local Pickup Plus.

### Additional Details

Tests were not running for me to begin with, as the test plugin included had a mismatching framework version in the constant than the one actually found in the namespace of each test plugin file. Also the `readme.md`, file was missing some instruction, which I have added here for the convenience of others too.

#### Test suites

Codeception/WPBrowser supports multiple suites. What is a suite? Just a collection of tests for which the testing platform will launch different modules and do different things. Other than that the nomenclature around them is just convetional. In fact, the `integration` suite, if I am not mistaken, is introduced by WPBrowser alone ([see here](https://codeception.com/docs/01-Introduction) where Codeception mentions only three). This has to do with the nature of the project being tested. 

WPBrowser documentation has an interesting overview about the [different levels of testing](https://wpbrowser.wptestkit.dev/levels-of-testing) matched to the test suites it offers and bootstraps for us when we first initialize it.

Regardless of how these are called, I would argue that we are interested in testing 2 things mainly:

1. That a product works as expected by the user, who is unaware of the product internals, following definitions and acceptance criteria defined by product management.
1. That a product internal logic works as expected by us, the people who need to extend it and write more code for it, as well as 3rd party developers who might use its APIs.

Therefore, I'd argue we could keep just 2 suites for these 2 scopes; currently in `master` are defined `acceptance`, `unit`, `functional` and `integration`. These are all nice approaches, but I'd argue that we struggle already with basic testing coverage, I don't think we'd ever achieve full coverage under 4 different umbrellas. Moreover, we'd risk dispersion and confusion when adding more tests. So, I'd propose we keep only 2:

1. **User behavior** ("acceptance"): tests are run into a browser instance simulating user interaction with UI: internal logic is not as important as expected results as perceived by the user)
1. **Logic and return types** ("integration"): tests are run with a WordPress DB instance to test how our methods interact with the WordPress database, as well as with WordPress core functions. 

What we normally consider unit tests could be bundled along with integration tests (as in within the same suite): when possible we test these in isolation (they will just evaluate asserts); for tests that need to set something in database, the integration suite supplied by WPBrowser provides the means to know whether a certain information was stored in WPDB. In this way we don't need to worry about mocking WordPress entities, etc. 

Another convention we could adopt is to name each integration `_Test` file after its corresponding class in the project, so they are easy to identify. Same goes for methods being tested. The directory structure can also follow the one from the actual project.

At the moment only the integration suite is running properly  since we don't have any acceptance tests - which we can ignore for now.

### Setup

Run `composer install` if you haven't already. Also the first time you clone this repository and wanna run tests, you should probably run `vendor/bin/codecept build`. (Same should be done if you add modules, helper change Codeception configuration, etc.)

Also, you need to make sure you have a database instance running locally. This will probably be zapped by Codeception/WPBrowser every time tests are run, so don't use one where you have an active WordPress installation used for other purposes. I set up mine on the machine where I will run my tests like so:

```shell
 sudo mysql -e "CREATE DATABASE IF NOT EXISTS framework_tests"  -uroot
```

where `framework_tests` should match the configuration in `.env`

Then, finally, run `vendor/bin/codecept run integration` to run your integration tests.


unfulvio:
@ChaseWiseman I've updated this to retain the Unit suite as in `master`. I still haven't changed my mind as to keep `acceptance` and `integration` only.

Some reasons:

#### Organization

I'd propose we'd organize integration and unit tests following the same project structure.

For example if we have a class inside `Admin\Settings.php` then we would create a file that covers integration & unit tests in `tests\integration\Admin\Settings.php`. 

If the `Settings.php` class contains a `get_setting_field()` method, then we would create a `test_get_setting_field()` method that tests the method. 

Having both an integration and a unit suite would duplicate this structure.

I'd also recommend to place a single line phpdoc on top of the method:

```php

use \SkyVerge\WooCommerce\PluginFramework\v5_5_1 as Framework;
use My_Plugin\Namespace\Admin;

/** @see Settings */
class Settings extends \Codeception\TestCase\WPTestCase {

	/** @see Admin\Settings::get_setting_field() */
	public function test_get_setting_field() {

		// run my assertions...
	}

}
```

#### No need for data providers

With integration tests using `WPTestCase` we don't need extra methods that define data providers (which I personally find also a bit confusing: I would rather define a key-value array containing arguments passed and expected results than an indexed array of values). The usage of special phpdoc tag like `dataProvider` is also not as intuitive. Test methods would never use any arguments (except protected / private helper methods, should we need any). 

#### Simpler setup

No need to include files directly with `__before()`, no need to check abspath. 

#### Minimal impact

As you've noticed too, the vast majority of tests we are going to run is going to be integration tests requiring the `WPLoader` module. So any speed impact should be also negligible. 

#### Using traits for helper methods across integration tests

Not specifically related to unit tests, but I should mention because it's part of the PR.

Memberships uses these a lot to set up factories for common objects. I chose traits because the editor's static analyzer understands very clearly what are the available methods included by that trait. Using helper modules from Codeception may not be as straightforward because the editor does not understand those.

ChaseWiseman:
@unfulvio thanks for updating!

>Tests were not running for me to begin with, as the test plugin included had a mismatching framework version in the constant than the one actually found in the namespace of each test plugin file.

Just a note, this wouldn't be the source of any fatal. The loader class constant is only used by Sake (for normal plugins) when sanity checking the FW version against what's loaded before deploy, as well as for `class_exists()` checking before including the FW files.

>I'd propose we'd organize integration and unit tests following the same project structure.

Agreed, and changing the test class name to match `SV_WC_Plugin` here, for example, lgtm.

That shouldn't prevent us from continuing with the unit test suite, however. We could take it case-by-case, as a class that requires integration tests probably wouldn't need _both_ unit and integration tests. Whereas a class like the country converter or string helpers need no integration testing, so unit is sufficient (even if they contain methods like `__()`, which are fine for mocking).

>No need for data providers

Data providers provide a cleaner way to deal with multiple iterations of a test, and are much more informative about failures. Consider the following output, generated by purposefully failing [only one of the assertions](https://github.com/skyverge/wc-plugin-framework/pull/368/files#diff-870ea40fdcb9d4921f9ea311108435cdR126) in `test_is_plugin_active`, and the difference between using a provider vs. looping directly within a test method:

#### Via loop (this branch)
```
✖ SV_WC_Plugin_Test: Test_is_plugin_active (0.00s)

There was 1 failure:

---------
1) SV_WC_Plugin_Test: Test_is_plugin_active
```

#### Via provider
```
✔ PluginTest: Test_is_plugin_active | #0 (0.00s)
✖ PluginTest: Test_is_plugin_active | #1 (0.00s)
✔ PluginTest: Test_is_plugin_active | #2 (0.00s)

There was 1 failure:

---------
1) PluginTest: Test_is_plugin_active | #1
```

When the looped version fails on any of the cases, it just fails without information into which case didn't pass. It also halts the process and the subsequent cases don't run. When using the provider, we know exactly which set of data failed, as well as still run the other sets of data to know that they pass. Additionally, whenever possible the `test_` methods themselves should not contain logic and ideally only one assertion.

>No need to include files directly with __before(), no need to check abspath.

The abspath check can just as easily go into the unit test suite's bootstrap file, then it would only be needed once. As could the includes, though I think having them at the top of the test class is reasonable and informative.

A couple of other qs for clarification on changes in this PR:
- What's the reason for removing [these helper classes](https://github.com/skyverge/wc-plugin-framework/pull/368/files#diff-7f8d746ddf4f35bfbe835df6f24085ebL7) for each test suite? I understand they're currently empty, but might be useful in the future
- What purpose does [WPLoader](https://github.com/skyverge/wc-plugin-framework/pull/368/files#diff-686b46a651b3297499b18e6082400ac1R38) serve in the acceptance test context? I had thought that was only used for integration tests, but what are the acceptance-level benefits?
- I don't think any of the tests files need to live under an additional `/woocommerce` dir - though it's named poorly, that dir is essentially `/src` in this context and there won't be any tests for files _outside_ that directory.

unfulvio:
hey @ChaseWiseman thanks for the feedback

> Data providers provide a cleaner way to deal with multiple iterations of a test, and are much more informative about failures. Consider the following output (non-essential content removed), generated by purposefully failing only one of the assertions in test_is_plugin_active, and the difference between using a provider vs. looping directly within a test method:

Also tests in a foreach loop as in the code I've provided in the PR are able to tell why a test they failed, by returning a sentence like `tests/integration/woocommerce/SV_WC_Plugin_Test.php:test_is_plugin_active
Failed asserting that true matches expected false` - I just double checked by deliberately fail an integration test; it would even point to the line of code where this failed. Would not this be the same for unit tests that adopt the same approach? Moreover I am not sure if data providers can be used in integration tests though, since they are run differently. 


> - What's the reason for removing [these helper classes](https://github.com/skyverge/wc-plugin-framework/pull/368/files#diff-7f8d746ddf4f35bfbe835df6f24085ebL7) for each test suite? I understand they're currently empty, but might be useful in the future

I've explained here:

> #### Using traits for helper methods across integration tests
> Not specifically related to unit tests, but I should mention because it's part of the PR.
> Memberships uses these a lot to set up factories for common objects. I chose traits because the editor's static analyzer understands very clearly what are the available methods included by that trait. Using helper modules from Codeception may not be as straightforward because the editor does not understand those.

Basically what I noticed, at least in my setup, if I included one such helper, when I actually write tests that need to use some methods provided by the helper, they are never recognized by the IDE. The editor is unable to understand the code I'm writing, unable to jump through code, recognize types and so on. 

I managed to fix this with traits, which are recognized by the IDE since they're part of the class, and you only need to load them when necessary.

I don't mind using the helper modules if we manage to fix that. I came up with this idea as a workaround to move forward when writing tests in Memberships/LPP. In the end they just provide a means to an end. 

Either approach we take might change if we decide to roll our own Codeception modules as a reusable library. But that will greatly depend on how we'd move forward with the test suite, and understanding how it would work out. I'm not quite sure on any of this yet. In the end, I predict if we're going to take any structured/unified approach to our plugins testing, a lot of changes will happen no matter what it is decided now. Well hopefully except the actual tests :) 

> - What purpose does [WPLoader](https://github.com/skyverge/wc-plugin-framework/pull/368/files#diff-686b46a651b3297499b18e6082400ac1R38) serve in the acceptance test context? I had thought that was only used for integration tests, but what are the acceptance-level benefits?

I just copied files from Memberships. IIRC I might have added that because it helped me writing some helpers in the factories or for handling something in the REST cases. As explained here:

https://github.com/lucatume/wp-browser/blob/master/docs/modules/WPLoader.md

> the module can be additionally used in acceptance and functional tests to acccess WordPress code in the tests context

basically you can use (almost) any of the modules in each suite; even asserts are used in REST tests in memberships to evaluate some values in the json response

> - I don't think any of the tests files need to live under an additional `/woocommerce` dir - though it's named poorly, that dir is essentially `/src` in this context and there won't be any tests for files _outside_ that directory.

normally I would use `/plugin`, I've used `woocommerce` because it's not a plugin and because it matches the dir in the project; I use a dir to keep it separated from `_helpers` and leave in the root of the tests suite only `_bootstrap.php`

ChaseWiseman:
@unfulvio 

>Also tests in a foreach loop as in the code I've provided in the PR are able to tell why a test they failed, by returning a sentence like tests/integration/woocommerce/SV_WC_Plugin_Test.php:test_is_plugin_active Failed asserting that true matches expected false

Yes, they do list the expected vs. actual that failed, but taking this particular test as an example there are multiple expected `false`, so which one failed?

>I just double checked by deliberately fail an integration test; it would even point to the line of code where this failed.

This is the line that contains the assert only, which will be the same for each test you're looping. So you could have a stack of 20 cases to loop through, but any one of them would list the same line 133 as the failure point. Using providers will list the line in the same way, but they also provide the exact test # that failed, as well as allow subsequent tests in the stack to run.

>Moreover I am not sure if data providers can be used in integration tests though, since they are run differently.

Yep, they're already being used in `master` successfully [here](https://github.com/skyverge/wc-plugin-framework/blob/master/tests/integration/PluginTest.php#L168-L193) 👍 

>Basically what I noticed, at least in my setup, if I included one such helper, when I actually write tests that need to use some methods provided by the helper, they are never recognized by the IDE. The editor is unable to understand the code I'm writing, unable to jump through code, recognize types and so on.

Yeah I see what you mean. Looks like we're supposed to run `codecept build` after adding a method to the helper, as outlined [here](https://codeception.com/docs/06-ModulesAndHelpers#helpers). That adds it to the classes/docs in `_generated`, and it becomes available via the `$tester` object.

unfulvio:
those are good points, let's move forward with the test added in `v5.6.0` then we will get back to this anyway when dealing with a more comprehensive testing approach later on -- thanks for all the comments and the discussion here

PR #369, opened 2019-12-12
Prevent js error on my account > payment methods screen

simonporter007:
# Summary

This PR adds a check to the `render_js` function to make sure there are existing tokens, therefore the front end class has been loaded, before attempting to initialize said class.

### QA

- [x] Navigating to my-account > payment methods with no previously saved tokens should not throw a JavaScript ReferenceError.
- [x] Navigating to my-account > payment methods with a saved token should still display previously saved cards.

### DIFF
--- a/woocommerce/changelog.txt
+++ b/woocommerce/changelog.txt
@@ -2,6 +2,7 @@
 
 2019.nn.nn - version 5.6.0-dev.1
  * Tweak - Migrate payment tokens to be compatible with WC core payment tokens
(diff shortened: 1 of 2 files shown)


wvega:
@unfulvio it seems we don't have an `.editorconfig` file on this repo. I thought you added one recently, but maybe we never merged it?

unfulvio:
@wvega ¯\\_(ツ)_/¯ did it now copying from `wc-plugins` https://github.com/skyverge/wc-plugin-framework/commit/7a075071172984be36baa88a4e149cd3fcfaebd1

PR #370, opened 2019-12-13
Implement SV_WC_Payment_Gateway_Payment_Token::save()

deivamagalhaes:
# Summary

This PR adds a `save()` method to the `SV_WC_Payment_Gateway_Payment_Token` class.

#### Setup

- Saved a payment method in a configured gateway
- Set up the gateway to use this version of the framework (I've symlinked it in the `vendor/skyverge` folder)
- Temporarily update `SV_WC_Payment_Gateway_Payment_Tokens_Handler::build_token` to set all the properties needed and save the token:
```
$data['type'] = 'credit_card';
$data['gateway_id'] = $this->get_gateway()->get_id();
$data['user_id'] = 8;

$token = new SV_WC_Payment_Gateway_Payment_Token( $token, $data );

$token->save();

return $token;
```
- Add `gateway_id` and `user_id` to the `props` property in `SV_WC_Payment_Gateway_Payment_Token` (if #364 and #365 are not merged yet)

#### Steps

1. Add a new payment method in your payment gateway using this branch of the framework
    - [x] The payment token in user meta is updated with the `migrated` key set to true/1
    - [x] The payment token is saved in the `wp_woocommerce_payment_tokens` and its meta is saved in `wp_woocommerce_payment_tokenmeta`
1. Refresh the page
    - [x] The payment token is not duplicated in `wp_woocommerce_payment_tokens`

### DIFF
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -666,7 +666,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 				if ( false !== $core_key ) {
 					$token->set_props( [ $core_key => $value ] );
 				} else {
-					$token->add_meta_data( $key, $value, true );
+					$token->update_meta_data( $key, $value, true );
 				}
 			}
 
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -538,6 +538,19 @@ class SV_WC_Payment_Gateway_Payment_Token {
 	}
 
 
+	/**
+	 * Stores the token data in the database.
+	 *
+	 * Stores the token data as a Woocommerce payment token, in the `wp_woocommerce_payment_tokens` table,
+	 * with meta data in the `wp_woocommerce_payment_tokenmeta` table.
+	 *
+	 * @since 5.6.0-dev.1
+	 */
+	public function save() {
+
+	}
+
+
 }
 
 
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -593,6 +593,10 @@ class SV_WC_Payment_Gateway_Payment_Token {
 
 				foreach ( $this->data as $key => $value ) {
 
+					if ( 'exp_year' === $key && 2 === strlen( $value ) ) {
+						$value = '20' . $value;
+					}
+
 					$core_key = array_search( $key, $this->props );
 
 					if ( false !== $core_key ) {
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -563,6 +563,16 @@ class SV_WC_Payment_Gateway_Payment_Token {
 					break;
 				}
 			}
+
+			if ( empty( $token ) ) {
+
+				// instantiate a new token
+				if ( $this->is_credit_card() ) {
+					$token = new \WC_Payment_Token_CC();
+				} elseif ( $this->is_echeck() ) {
+					$token = new \WC_Payment_Token_ECheck();
+				}
+			}
 		}
 	}
 
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -236,7 +236,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 	 */
 	public function is_credit_card() {
 
-		return 'credit_card' == $this->data['type'];
+		return 'credit_card' === $this->data['type'];
 	}
 
 
@@ -628,9 +628,9 @@ class SV_WC_Payment_Gateway_Payment_Token {
 
 		$token = $this->get_woocommerce_payment_token();
 
-		if ( empty( $token ) ) {
+		if ( ! $token instanceof \WC_Payment_Token ) {
 
-			// instantiate a new token
+			// instantiate a new token: if it's not a credit card, we default to echeck, so there's always an instance
 			if ( $this->is_credit_card() ) {
 				$token = new \WC_Payment_Token_CC();
 			} elseif ( $this->is_echeck() ) {
@@ -649,29 +649,26 @@ class SV_WC_Payment_Gateway_Payment_Token {
 			}
 		}
 
-		if ( ! empty( $token ) ) {
+		$token->set_token( $this->get_id() );
 
-			$token->set_token( $this->get_id() );
+		foreach ( $this->data as $key => $value ) {
 
-			foreach ( $this->data as $key => $value ) {
-
-				// prefix the expiration year if needed (WC_Payment_Token requires it to be 4 digits long)
-				// TODO: figure out how to handle cards expiring before 2000 {DM 2019-12-13}
-				if ( 'exp_year' === $key && 2 === strlen( $value ) ) {
-					$value = '20' . $value;
-				}
-
-				$core_key = array_search( $key, $this->props );
-
-				if ( false !== $core_key ) {
-					$token->set_props( [ $core_key => $value ] );
-				} else {
-					$token->update_meta_data( $key, $value, true );
-				}
+			// prefix the expiration year if needed (WC_Payment_Token requires it to be 4 digits long)
+			// TODO: figure out how to handle cards expiring before 2000 {DM 2019-12-13}
+			if ( 'exp_year' === $key && 2 === strlen( $value ) ) {
+				$value = '20' . $value;
 			}
 
-			$token->save();
+			$core_key = array_search( $key, $this->props, false );
+
+			if ( false !== $core_key ) {
+				$token->set_props( [ $core_key => $value ] );
+			} else {
+				$token->update_meta_data( $key, $value, true );
+			}
 		}
+
+		$token->save();
 	}
 
 
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -558,7 +558,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 
 			foreach ( $saved_tokens as $saved_token ) {
 
-				if ( $saved_token->get_id() === $this->get_id() ) {
+				if ( $saved_token->get_token() === $this->get_id() ) {
 					$token = $saved_token;
 					break;
 				}
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -575,6 +575,16 @@ class SV_WC_Payment_Gateway_Payment_Token {
 
 				// mark the token as migrated
 				$this->data['migrated'] = true;
+
+				// update legacy token in user meta data
+				$gateways   = WC()->payment_gateways->payment_gateways();
+				// TODO: use `get_gateway_id`, once merged {DM 2019-12-12}
+				$gateway_id = $this->data['gateway_id'];
+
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+	 *
+	 * @since 5.6.0-dev.1
+	 */
+	public function save() {
+
+		if ( $this->token instanceof \WC_Payment_Token ) {
+			$token = $this->token;
+		} else {
+
+			// see if there is already a token with this ID saved for this customer and gateway
```
deivamagalhaes: I was thinking maybe this part (the variable assignment and the foreach) could be moved to `get_woocommerce_payment_token`

unfulvio: yeah actually I was thinking that too, because there's a couple of instances (save, delete) where we would need this - sounds good to me

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+				}
+			}
+
+			if ( ! empty( $token ) ) {
+
+				$token->set_token( $this->get_id() );
+
+				foreach ( $this->data as $key => $value ) {
+
+					if ( 'exp_year' === $key && 2 === strlen( $value ) ) {
```
unfulvio: let's make a note here why this is necessary, as you mentioned on Slack - out of curiosity I checked what's the maximum duration of a card, and some 50+ yr expiry cards do exist (...I wonder, right?)

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+						$value = '20' . $value;
+					}
+
+					$core_key = array_search( $key, $this->props );
+
+					if ( false !== $core_key ) {
+						$token->set_props( [ $core_key => $value ] );
+					} else {
+						// metadata
```
unfulvio: Comment seems unnecessary when you have a method that says `add_meta_data()` :wink: 

More importantly, should this be `add_meta_data()` or `update_meta_data()`? Because the `save()` method could be used to update a token as well - we don't want skips for existing data (since `unique` is set to `true`).

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -584,6 +584,35 @@ public function update_tokens( $user_id, $tokens, $environment_id = null ) {
...
+	 * @internal
+	 *
+	 * @since 5.6.0-dev.1
+	 *
+	 * @param int $user_id WP user ID
+	 * @param SV_WC_Payment_Gateway_Payment_Token $token token to update
+	 * @param string|null $environment_id optional environment ID, defaults to plugin current environment
+	 * @return string|int updated user meta ID
+	 */
+	public function update_legacy_token( $user_id, $token, $environment_id = null ) {
```
unfulvio: Howe about keeping this helper as private inside the `SV_WC_Payment_Gateway_Payment_Token` object? I see you are fetching the payment tokens handler from the gateway object. Perhaps you could delegate [this](https://github.com/skyverge/wc-plugin-framework/blob/50f57745f9d3cfee1570ed3c1bdb27f29a42d3db/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php#L584-L587) to the helper method itself?

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -584,6 +584,35 @@ public function update_tokens( $user_id, $tokens, $environment_id = null ) {
...
+
+		// default to current environment
+		if ( null === $environment_id ) {
+			$environment_id = $this->get_environment_id();
+		}
+
+		$legacy_tokens = get_user_meta( $user_id, $this->get_user_meta_name( $environment_id ), true );
```
unfulvio: Even if `$legacy_tokens` is `false` (default return value here), I believe PHP type juggling should be able to convert to array and let us set the legacy token as you do right below. However, I generally don't trust data coming from `get_user_meta()` since it can be of any type and I'd default `$legacy_tokens`  to array type if it's not one.

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+	 *
+	 * Stores the token data as a Woocommerce payment token, in the `wp_woocommerce_payment_tokens` table,
+	 * with meta data in the `wp_woocommerce_payment_tokenmeta` table.
+	 *
+	 * @since 5.6.0-dev.1
+	 */
+	public function save() {
+
+		if ( $this->token instanceof \WC_Payment_Token ) {
+			$token = $this->token;
```
unfulvio: this instance of `$token` will be unused: `::save()` will never be called on it (`$token->save()` is currently nested in another conditional)

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+
+				$token->set_token( $this->get_id() );
+
+				foreach ( $this->data as $key => $value ) {
+
+					if ( 'exp_year' === $key && 2 === strlen( $value ) ) {
+						$value = '20' . $value;
+					}
+
+					$core_key = array_search( $key, $this->props );
```
unfulvio:
```suggestion
					$core_key = array_search( $key, $this->props, false );
```

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+					$token = new \WC_Payment_Token_CC();
+				} elseif ( $this->is_echeck() ) {
+					$token = new \WC_Payment_Token_ECheck();
+				}
+
+				// mark the token as migrated
+				$this->data['migrated'] = true;
+
+				// update legacy token in user meta data
+				$gateways   = WC()->payment_gateways->payment_gateways();
```
unfulvio:
```suggestion
				$gateways   = WC()->payment_gateways()->payment_gateways();
```

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+				// TODO: use `get_gateway_id`, once merged {DM 2019-12-12}
+				$gateway_id = $this->data['gateway_id'];
+
+				if ( ! empty( $gateway_id ) && ! empty( $gateway = $gateways[ $gateway_id ] ) ) {
```
unfulvio: uhm maybe should we check if `get_payment_tokens_handler()` exists for that gateway for sanity?

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -538,6 +538,81 @@ private function read( \WC_Payment_Token $core_token ) {
...
+
+					if ( false !== $core_key ) {
+						$token->set_props( [ $core_key => $value ] );
+					} else {
+						// metadata
+						$token->add_meta_data( $key, $value, true );
+					}
+				}
+
+				$token->save();
```
unfulvio: this should be moved further below in case `$token` is already set on the object so it can be saved

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -584,6 +584,35 @@ public function update_tokens( $user_id, $tokens, $environment_id = null ) {
...
+	 * Updates a single legacy token in user meta.
+	 *
+	 * @internal
+	 *
+	 * @since 5.6.0-dev.1
+	 *
+	 * @param int $user_id WP user ID
+	 * @param SV_WC_Payment_Gateway_Payment_Token $token token to update
+	 * @param string|null $environment_id optional environment ID, defaults to plugin current environment
+	 * @return string|int updated user meta ID
```
unfulvio: `update_user_meta()` can return `int|bool` and not `string` IIRC 

does this method need to return anything though?

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -584,6 +584,35 @@ public function update_tokens( $user_id, $tokens, $environment_id = null ) {
 	}
 
 
+	/**
+	 * Updates a single legacy token in user meta.
+	 *
+	 * @internal
```
unfulvio: `@internal` tag not needed if the method is accessed from elsewhere outside the class - let's keep this for hook callbacks

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
unfulvio: see another note below about the method being called here, maybe this logic can be kept inside the helper itself

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
deivamagalhaes: I believe this was fixed when I moved the search code to `get_woocommerce_payment_token`.

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
deivamagalhaes: Isn't the default null (not true)?

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
deivamagalhaes: Can you please clarify why?

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
deivamagalhaes: This method is heavily "inspired" in the existing `update_tokens`, so I thought I could keep the return type, for consistency.

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
deivamagalhaes: This helper method uses `get_environment_id` and `get_user_meta_name` from the handler and is very similar to the existing `update_tokens` (that will be changed soon). I do believe it makes sense for it to live on the handler... Do you believe it would be better for it to live in SV_WC_Payment_Gateway_Payment_Token object and retrieve the handler to call these methods?

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
unfulvio: WC does not define a `payment_gateways` property; instead it uses a magic method `__get()` to assign it to `payment_gateways()` method. It is harder for editors and static analyzers understand this pattern and may mark the property as non-existing; also the return type would not be clear. Instead, calling the method directly is more transparent. It's easier to point where  a problem might be when properties or methods are declared explicitly than dynamically via `__call` and `__get` in my opinion. Since you are on PTO, if you don't mind I would take the initiative to change this before merging. If you want discuss this later, please do so.

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
unfulvio: fair enough, let's keep it as you propose - we can always review this later if we feel to  -- thanks

PR #371, opened 2019-12-16
Implement SV_WC_Payment_Gateway_Payment_Token::delete()

unfulvio:
# Summary

This PR adds a `delete()` method to the `SV_WC_Payment_Gateway_Payment_Token` class.

#### Setup

Same setup as #370 

- Have a gateway configured with some test order placed
- Set up the gateway to use this version of the framework
- Temporarily update `SV_WC_Payment_Gateway_Payment_Tokens_Handler::build_token()` to set all the properties needed and save a token first:
```php
$data['type'] = 'credit_card';
$data['gateway_id'] = $this->get_gateway()->get_id();
$data['user_id'] = 8;

$token = new Framework\SV_WC_Payment_Gateway_Payment_Token( $token, $data );

$token->save();
```

This should give you a token as in #370. 

#### Steps

Verify you have some token in db, saved as in `Setup` above.

Then, you may use the following to delete a token:

```php

// retrieve the token you have saved before
$core_token = new \WC_Payment_Token_CC( $token_identifier );

// instantiate our token 
$fw_token = new Framework\SV_WC_Payment_Gateway_Payment_Token( $user_id, core_token );

// delete the token:
$fw_token->delete();
```

The token now should be deleted from database.

Repeat the steps above with a legacy token that was saved and migrated: in this instance, both the legacy data from the corresponding WordPress user meta and the token entry in WooCommerce tables should be wiped.

### DIFF
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -673,7 +673,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 
 
 	/**
-	 * Deletes a token from the database.
+	 * Deletes the token from the database.
 	 *
 	 * Also deletes a token from the legacy user meta.
 	 * @see \WC_Payment_Token::delete()
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -689,20 +689,48 @@ class SV_WC_Payment_Gateway_Payment_Token {
 
 		$deleted = false;
 
+		// delete the core token from WooCommerce tables
 		if ( $token = $this->get_woocommerce_payment_token() ) {
 			$deleted = $token->delete( $force_delete );
 		}
 
-		// delete legacy token in user meta data
-		$gateways   = WC()->payment_gateways()->payment_gateways();
-		$gateway_id = $this->get_gateway_id();
-
-		if ( ! empty( $gateway_id ) && ! empty( $gateway = $gateways[ $gateway_id ] ) ) {
-			/** @see SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_legacy_token() */
-			$gateway->get_payment_tokens_handler()->delete_legacy_token( $this->get_user_id(), $this );
+		// delete legacy token in WordPress user meta table
+		if ( $tokens_handler = $this->get_tokens_handler() ) {
+			$tokens_handler->delete_legacy_token( $this->get_user_id(), $this );
 		}
 
 		return $deleted;
+	}
+
+
+	/**
+	 * Gets the gateway tokens handler.
+	 *
+	 * @since 5.6.0-dev.1
+	 *
+	 * @return SV_WC_Payment_Gateway_Payment_Tokens_Handler|null
+	 */
+	protected function get_tokens_handler() {
+
+		$handler    = null;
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -672,6 +672,39 @@ public function save() {
...
+	 */
+	public function delete( $force_delete = false ) {
+
+		$deleted = false;
+
+		if ( $token = $this->get_woocommerce_payment_token() ) {
+			$deleted = $token->delete( $force_delete );
+		}
+
+		// delete legacy token in user meta data
```
unfulvio: this follows a pattern as in `save()` method, which for now utilizes a tokens handler method to perform operations with legacy tokens

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -672,6 +672,39 @@ public function save() {
...
+		$gateways   = WC()->payment_gateways()->payment_gateways();
+		$gateway_id = $this->get_gateway_id();
+
+		if ( ! empty( $gateway_id ) && ! empty( $gateway = $gateways[ $gateway_id ] ) ) {
+			/** @see SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_legacy_token() */
+			$gateway->get_payment_tokens_handler()->delete_legacy_token( $this->get_user_id(), $this );
```
unfulvio: the environment could be passed as we implement a method in the token to get the environment, I'll make a note in the related story

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -672,6 +672,39 @@ public function save() {
 	}
 
 
+	/**
+	 * Deletes a token from the database.
```
ChaseWiseman:
```suggestion
	 * Deletes the token from the database.
```

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -672,6 +672,39 @@ public function save() {
...
+	 *
+	 * Also deletes a token from the legacy user meta.
```
ChaseWiseman:
```suggestion
	 * Also deletes the token from the legacy user meta.
	 *
```

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -672,6 +672,39 @@ public function save() {
...
+
+		if ( $token = $this->get_woocommerce_payment_token() ) {
+			$deleted = $token->delete( $force_delete );
+		}
+
+		// delete legacy token in user meta data
+		$gateways   = WC()->payment_gateways()->payment_gateways();
+		$gateway_id = $this->get_gateway_id();
+
+		if ( ! empty( $gateway_id ) && ! empty( $gateway = $gateways[ $gateway_id ] ) ) {
```
ChaseWiseman: we should break this var assignment out of the conditional for readability. Also, should check that the gateway is an instance of `\SV_WC_Payment_Gateway` before calling the tokens handler, just in case.

unfulvio: Ok I will have to update both this bit and the one in `save()` since they are quite the same

PR #372, opened 2019-12-16
Fix verification in SV_WC_Payment_Gateway_Apple_Pay::process_payment()

wvega:
# Summary

This PR fixes `SV_WC_Payment_Gateway_Apple_Pay::process_payment()` to throw an exception if the processing gateway doesn't return a successful response.

Additionally, it updates `SV_WC_Payment_Gateway_Direct::process_payment()` to cover for and edge case in which `SV_WC_Payment_Gateway_Direct::do_transaction()` fails to create a transaction, but returns `false` instead of throwing an exception.

## QA

This bug is an edge case I discovered while performing some Apple Pay tests for [HS 111556](https://secure.helpscout.net/conversation/953413936/111556/#thread-2912230502) and not having a device that supports Apple Pay. As a results the steps to test are a bit hacky.

### Setup

Activate WooCommerce Authorize.Net Gateway and enable Apple Pay. You can use a fake merchant ID and certificate as no real transactions will be processed.

### Using the version of the framework included in the extension

1. Go to a product page and add the product to the cart
1. Execute the following on the browsers console:

	    sv_wc_apple_pay_handler.process_authorization( {} )

	- [x] The response for the associated Ajax request is `{"success":true,"data":[]}`

### Using this branch of the framework

* Symlink or replace the copy of `woocommerce/payment-gateway/apple-pay/class-sv-wc-payment-gateway-apple-pay.php` in the extension's vendor directory with a copy of the file in this branch
* Symlink or replace the copy of `woocommerce/payment-gateway/class-sv-wc-payment-gateway-direct.php` in the extension's vendor directory with a copy of the file in this branch

Symlinking the entire project causes scripts not be loaded correctly because URLs are not properly generated (at least on my setup).

1. Go to a product page and add the product to the cart
1. Execute the following on the browsers console:

	    sv_wc_apple_pay_handler.process_authorization( {} )

	- [x] The response for the associated Ajax request is `{"success":false,"data":{"message":"Gateway processing error.","code":0}}`

## Before merge

### DIFF
--- a/woocommerce/changelog.txt
+++ b/woocommerce/changelog.txt
@@ -1,5 +1,10 @@
 *** SkyVerge WooCommerce Plugin Framework Changelog ***
 
+2019.nn.nn - version 5.6.0-dev.1
+
+* Fix - `SV_WC_Payment_Gateway_Apple_Pay::process_payment()` now throws an exception if the result returned by the processing gateway doesn't indicate whether the transaction was successful or not
+* Fix - Update `SV_WC_Payment_Gateway_Direct::process_payment()` to cover for and edge case in which `SV_WC_Payment_Gateway_Direct::do_transaction()` fails without throwing an exception
+
 2019.11.14 - version 5.5.1
  * Tweak - Refactor Apple Pay order creation to support the same filters and actions that are fired during regular checkout
  * Tweak - Allow multiple old hooks to be mapped to a single new one via the hook deprecator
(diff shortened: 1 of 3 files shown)


wvega:
>Looks good to me - should we add a changelog entry for this?

@unfulvio we probably should. I added changelog entries in d151125 using `5.6.0-dev.1` for the version. Not sure if we are going to start using the `x.y.z` convention now or how.

unfulvio:
also, this bit from the PR template might no longer apply if merged directly to master:

> - [ ] I have copied all UI Changes and QA items listed above into the base release branch PR description

PR #373, opened 2019-12-18
Implement SV_WC_Payment_Gateway_Payment_Token::get_environment()

wvega:
# Summary

This PR implements the `SV_WC_Payment_Gateway_Payment_Token::get_environment()` and updates `SV_WC_Payment_Gateway_Payment_Token::delete()` to pass the token enviroment to `SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_legacy_token()`.

## Details

This PR does not update `SV_WC_Payment_Gateway_Payment_Token::save()` to pass the token environment because we currently only update the legacy token when there is no `WC_Payment_Token` object, in which case there is no metadata to retrieve.

We may need to start taking the enviroment in the constructor, as we started doing with `gateway_id`.

## QA

- [x] SV_WC_Payment_Gateway_Payment_Token_Test::test_get_environment pass
- [x] SV_WC_Payment_Gateway_Payment_Token_Test::test_delete_passes_the_token_environment pass

## Before merge

### DIFF
--- a/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
+++ b/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
@@ -405,7 +405,7 @@ class SV_WC_Payment_Gateway_Payment_Token_Test extends \Codeception\TestCase\WPT
 
 		$token = new Framework\SV_WC_Payment_Gateway_Payment_Token( '12345', $woocommerce_token );
 
-		$this->assertSame( $expected_enviroment, $token->get_environment() );
+		$this->assertSame( $expected_environment, $token->get_environment() );
 	}
 
 
--- a/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
+++ b/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
@@ -397,7 +397,7 @@ class SV_WC_Payment_Gateway_Payment_Token_Test extends \Codeception\TestCase\WPT
 	 *
 	 * @dataProvider provider_get_enviroment
 	 */
-	public function test_get_environment( $stored_environment, $expected_enviroment ) {
+	public function test_get_environment( $stored_environment, $expected_environment ) {
 
 		$woocommerce_token = $this->get_new_woocommerce_credit_card_token();
 
@@ -410,11 +410,11 @@ class SV_WC_Payment_Gateway_Payment_Token_Test extends \Codeception\TestCase\WPT
 
 
 	/**
-	 * Provides test data for test_get_enviroment().
+	 * Provides test data for test_get_environment().
 	 *
 	 * @return array
 	 */
-	public function provider_get_enviroment() {
+	public function provider_get_environment() {
 
 		return [
 			'metadata is set'  => [ 'test_environment', 'test_environment' ],
@@ -437,7 +437,7 @@ class SV_WC_Payment_Gateway_Payment_Token_Test extends \Codeception\TestCase\WPT
 		$tokens_handler = \Codeception\Stub::make(
 			Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::class,
 			[
-				// mock delete_legacy_token() to check that the token enviroment is passed as the third parameter
+				// mock delete_legacy_token() to check that the token environment is passed as the third parameter
 				'delete_legacy_token' => \Codeception\Stub\Expected::once(
 					function( $user_id, $token, $environment_id ) use ( $environment ) {
 						$this->assertSame( $environment, $environment_id );
--- a/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
+++ b/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
@@ -418,6 +418,8 @@ class SV_WC_Payment_Gateway_Payment_Token_Test extends \Codeception\TestCase\WPT
 
 		return [
 			'metadata is set'  => [ 'test_environment', 'test_environment' ],
+			'empty metadata'   => [ '', null ],
+			'metadata not set' => [ null, null ],
 		];
 	}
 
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -540,6 +540,22 @@ public function set_billing_hash( $value ) {
...
 
 
+	/**
+	 * Gets the gateway environment that this token is associated with.
+	 *
+	 * @since 5.6.0-dev.1
+	 *
+	 * @return string|null
+	 */
+	public function get_environment() {
```
unfulvio: hmm I thought we were going with the same route as the other setters to read from `$data` once `read()` has done what this method in this PR is doing here

True, we currently don't pass `$environment` since this is part of the tokens handler; it'd be valuable that we do so moving forward, wouldn't it?

So, hypothetically, this method could return:

```php

$environment = isset( $this->data['environment'] ) && is_string( $this->data['environment'] ) ? $this->data['environment'] : '';

if ( '' === $environment && ( $handler = $this->get_tokens_handler() ) ) {
	$environment = $handler->get_gateway()->get_environment();
}

return $environment;

```

Perhaps the `::build_token()` method in the handler could pass the current environment when using `$data` as an array, since it has context of the environment?

And also, perhaps the `get_environment()` method could fallback on the value currently set in the handler? Because it's essentially what's happening already in the context of this class when we use the method internally: if hypothetically returns empty, then the handler will still get the default defined on the handler itself. Since this method is public, it could have other uses though.

maybe @ChaseWiseman could offer a 2nd opinion here

@wvega was also suggesting we could have a setter too

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
```diff
@@ -392,6 +392,74 @@ public function test_set_billing_hash() {
...
+		$this->assertSame( $expected_enviroment, $token->get_environment() );
+	}
+
+
+	/**
+	 * Provides test data for test_get_enviroment().
+	 *
+	 * @return array
+	 */
+	public function provider_get_enviroment() {
```
unfulvio:
```suggestion
	public function provider_get_environment() {
```

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
```diff
@@ -392,6 +392,74 @@ public function test_set_billing_hash() {
...
+		$woocommerce_token->add_meta_data( 'environment', $stored_environment );
+
+		$token = new Framework\SV_WC_Payment_Gateway_Payment_Token( '12345', $woocommerce_token );
+
+		$this->assertSame( $expected_enviroment, $token->get_environment() );
+	}
+
+
+	/**
+	 * Provides test data for test_get_enviroment().
```
unfulvio:
```suggestion
	 * Provides test data for test_get_environment().
```

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
```diff
@@ -392,6 +392,74 @@ public function test_set_billing_hash() {
 	}
 
 
+	/**
+	 * @see Framework\SV_WC_Payment_Gateway_Payment_Token::get_environment()
+	 *
+	 * @dataProvider provider_get_enviroment
+	 */
+	public function test_get_environment( $stored_environment, $expected_enviroment ) {
```
unfulvio:
```suggestion
	public function test_get_environment( $stored_environment, $expected_environment ) {
```

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
```diff
@@ -392,6 +392,74 @@ public function test_set_billing_hash() {
...
+
+		$woocommerce_token = $this->get_new_woocommerce_credit_card_token();
+		$environment       = 'test_environment';
+
+		$woocommerce_token->add_meta_data( 'environment', $environment );
+
+		$tokens_handler = \Codeception\Stub::make(
+			Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::class,
+			[
+				// mock delete_legacy_token() to check that the token enviroment is passed as the third parameter
```
unfulvio:
```suggestion
				// mock delete_legacy_token() to check that the token environment is passed as the third parameter
```

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
wvega: >I thought we were going with the same route as the other setters to read from $data once read() has done what this method in this PR is doing here

I agree that we need to be able to read the environment when the framework token is instantiated.

I decided to follow the recommendations in the story because the tests and other modifications would still be useful even after changing `get_environment()` and adding a setter or taking the value in the constructor.

>And also, perhaps the get_environment() method could fallback on the value currently set in the handler?

I don't think we should have `get_environment()` fallback to `$this->get_tokens_handler()->get_gateway()->get_environment()`. One reason is because it currently forces us to use partial mocks for testing, which I think is better if we can avoid them. Also, I think we currently only pass the environment to the handler itself, which defaults to the gateway environment anyway, so long as we pass `null` whenever we don't know the environment.

The version of the method returns `null` as fallback, but using an empty string to maintain the return type works for me too. We can use `$this->get_environment() ?: null` to call `update_legacy_token()` and `delete_legacy_token()` and get the same behavior (or update those methods to handle empty strings the same way).

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Token_Test.php
wvega: 🤦🏻‍♂️ - Thank you for catching those typos.

PR #374, opened 2019-12-18
Update SV_WC_Payment_Gateway_Payment_Tokens_Handler::update_tokens()

ChaseWiseman:
# Summary

Migrates token updating to core handling.

## Details

Note: left transient/cache handling alone until we finalize the `::get_tokens()` approach.

### Setup

- Update the NETBilling gateway's composer version to `dev-ch24096/update-tokens`
- Activate and configure the gateway

### Steps

1. Place an order and save your credit card at checkout
    - [ ] The token is added to the database with the correct properties and meta

## Before merge

### DIFF
--- a/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
+++ b/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
@@ -27,6 +27,27 @@ class SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test extends \Codeception\Tes
 
 
 	/**
+	 * @see Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::update_tokens()
+	 */
+	public function test_update_tokens() {
+
+		$tokens = [
+			new Framework\SV_WC_Payment_Gateway_Payment_Token( '12345', [
+				'type' => 'credit_card',
+				'last_four' => '1111',
+				'exp_month' => '01',
+				'exp_year'  => '20',
+				'card_type' => 'visa',
+			] )
+		];
+
+		$this->get_handler()->update_tokens( 1, $tokens );
+
+		$this->assertIsArray( \WC_Payment_Tokens::get_customer_tokens( 1, 'test_gateway' ) );
+	}
+
+
+	/**
 	 * @see Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::get_user_meta_name()
 	 *
 	 * @dataProvider provider_get_user_meta_name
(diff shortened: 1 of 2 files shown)


tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
```diff
@@ -26,6 +26,27 @@ protected function _after() {
 	}
 
 
+	/**
+	 * @see Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::update_tokens()
+	 */
+	public function test_update_tokens() {
```
unfulvio: shouldn't we add tests about how data changes when it's updated? currently this only tests that data is retrieved, but there's no insight about its contents

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -613,8 +613,28 @@ public function update_tokens( $user_id, $tokens, $environment_id = null ) {
...
+			// TODO: set the environment ID {CW 2019-12-17}
+
+			$token->save();
+
+			if ( $core_token = $token->get_woocommerce_payment_token() ) {
+				$updated_tokens[] = $core_token->get_token();
+			}
+		}
+
+		return ! empty( $updated_tokens );
```
unfulvio: how about we return true/false based on the combined results of `\WC_Data::save()` ? that should return theoretically true/false if save was successful

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
ChaseWiseman: I figured any tests for ensuring different properties/meta are set should be testing the `::save()` method specifically. In the update tokens context, we're just testing that it was saved

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
ChaseWiseman: Added c3d078b7d67c7dddd81a7dad21deadf3fe313e35 to tweak the result a bit.

We'll actually probably end up removing this return and phasing out the use of update_tokens as we update the single token handler methods, as it doesn't hold much value either way. There's only one spot where its results are used, via `->add_token()`, so this return mimics that result, i.e. all of the tokens were or were not saved.

PR #375, opened 2019-12-18
Implement SV_WC_Payment_Gateway_Payment_Tokens_Handler::get_legacy_tokens()

ChaseWiseman:
# Summary

Adds a `get_legacy_tokens()` method to the token handler specifically for retrieving token objects from user meta.

## Details

Adds caching for the resulting legacy tokens.

## UI Changes

N/A

### Setup

- Install and configure NETBilling as-is
- Add a payment method to your account
- Update the FW composer version to `dev-ch24006/get_legacy_tokens`

### Steps

1. Add an item to your cart, visit Checkout
    - [x] Your saved method is available at checkout
1. Place the order
    - [x] The saved method is used successfully

## Before merge


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -590,6 +585,37 @@ public function get_tokens( $user_id, $args = array() ) {
...
+	 * @param null|string $environment_id desired environment ID
+	 * @return SV_WC_Payment_Gateway_Payment_Token[]
+	 */
+	public function get_legacy_tokens( $user_id, $environment_id = null ) {
+
+		if ( null === $environment_id ) {
+			$environment_id = $this->get_environment_id();
+		}
+
+		$this->legacy_tokens[ $environment_id ][ $user_id ] = [];
```
wvega: Based on the way `get_tokens()` works, I think we want to check whether we already have tokens on the cache and return those instead before this line.

ChaseWiseman:
@wvega thanks! Yeah I think we can re-assess any further cache changes after we've implemented the `get_tokens()` changes to use core tables, and this legacy getter in the core filters.

PR #376, opened 2019-12-18
Update SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_token()

wvega:
# Summary

This PR updates `SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_token()` to call `delete()` on the token object and stop calling `SV_WC_Payment_Gateway_Payment_Tokens_Handler::update_tokens()`.

## Details

`SV_WC_Payment_Gateway_Payment_Tokens_Handler::update_tokens()` also updates the internal cache, so in order to keep getting accurate results from `SV_WC_Payment_Gateway_Payment_Tokens_Handler::get_tokens()`, `delete_token()` now updates the internal cache and clear the user's transient too.

## QA

- [x] SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test::test_delete_token pass
- [x] SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test::test_delete_token_marks_another_token_as_deafult pass

## Before merge

### DIFF
--- a/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
+++ b/tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
@@ -58,6 +58,46 @@ class SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test extends \Codeception\Tes
 
 
 	/**
+	 * @see Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_token()
+	 */
+	public function test_delete_token_marks_another_token_as_deafult() {
+
+		$token_id         = '12345';
+		$default_token_id = '45678';
+
+		// store test tokens
+		$tokens = [
+			'12345' => new Framework\SV_WC_Payment_Gateway_Payment_Token( '12345', [
+				'type' => 'credit_card',
+				'last_four' => '1111',
+				'exp_month' => '01',
+				'exp_year'  => '20',
+				'card_type' => 'visa',
+				'default'   => true,
+			] ),
+			'45678' => new Framework\SV_WC_Payment_Gateway_Payment_Token( '45678', [
+				'type' => 'credit_card',
+				'last_four' => '2222',
+				'exp_month' => '01',
+				'exp_year'  => '20',
+				'card_type' => 'visa',
+			] ),
+		];
+
+		$this->get_handler()->update_tokens( 1, $tokens );
+
+		$this->get_handler()->delete_token( 1, $this->get_handler()->get_token( 1, $token_id ) );
+
+		$this->assertNull( $this->get_handler()->get_token( 1, $token_id ) );
+
+		$deafult_token = $this->get_handler()->get_token( 1, $default_token_id );
+
+		$this->assertInstanceOf( Framework\SV_WC_Payment_Gateway_Payment_Token::class, $deafult_token );
+		$this->assertTrue( $deafult_token->is_default() );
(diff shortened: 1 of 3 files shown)


tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
```diff
@@ -26,6 +26,81 @@ protected function _after() {
 	}
 
 
+	/**
+	 * @see Framework\SV_WC_Payment_Gateway_Payment_Tokens_Handler::delete_token()
+	 */
+	public function test_delete_token() {
```
ChaseWiseman: it doesn't look like this test contains any assertions

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
```diff
@@ -26,6 +26,81 @@ protected function _after() {
...
+				'exp_year'  => '20',
+				'card_type' => 'visa',
+			] ),
+		];
+
+		$this->get_handler()->update_tokens( 1, $tokens );
+
+		$this->get_handler()->delete_token( 1, $this->get_handler()->get_token( 1, $token_id ) );
+
+		$this->assertNull( $this->get_handler()->get_token( 1, $token_id ) );
```
ChaseWiseman: maybe this assert should be in `test_delete_token()`? If that's covered, then no need to retest a successful delete in this method, and can instead just focus on checking the default portion

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
```diff
@@ -26,6 +26,81 @@ protected function _after() {
...
+
+		$default_token = $this->get_handler()->get_token( 1, $default_token_id );
+
+		$this->assertInstanceOf( Framework\SV_WC_Payment_Gateway_Payment_Token::class, $default_token );
```
ChaseWiseman: can probably skip this assert as well, since it's really testing `get_token()`

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
wvega: The `\Codeception\Stub::make(...)` part causes the test to fail if `delete()` is **not** called on the token mock exactly once before the test ends.

However, I think I can write this test using an approach similar to the one used `test_delete_token_marks_another_token_as_deafult` which kind of aligns better with the integration nature of these tests. Then I can move some of those other asserts to this one as you suggested in another comment.

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
wvega: Removed in d2324fd9eaedfebb5273a93215e21a4b194c0176.

tests/integration/payment-gateway/SV_WC_Payment_Gateway_Payment_Tokens_Handler_Test.php
wvega: Added the assert to `test_delete_token()` in 4b99349 and removed it from this test in 468c997.

PR #381, opened 2019-12-20
Cast the token ID as string

unfulvio:
# Summary

This PR casts the token `$id` as a string when it's set in the `SV_WC_Payment_Gateway_Payment_Token` constructor. 

## QA

There is already a test that compares a string 

- [ ] SV_WC_Payment_Gateway_Payment_Token_Test::test_get_id() passes (updated test to ensure the id is always returned as a string)

## Before merge


asseti6:
Commit breakdowns lgtm!

PR #383, opened 2019-12-20
Remove token object legacy handling

ChaseWiseman:
# Summary

Removes references to the token handler from the base token object.

## Details

The way we were getting the gateway/handler is a bit brittle, and either way the token object shouldn't need to be aware of the token handler. Its only job is to hold the token data, and perform crud operations for itself.

Updated [CH 24264](https://app.clubhouse.io/skyverge/story/24264/implement-add-legacy-tokens-to-customer-payment-tokens-to-filter-woocommerce-get-customer-payment-tokens) to handle setting the `migrated` flag on legacy tokens that was previously done via the token object.

## UI Changes

N/A

## QA

- [x] All integration tests pass

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
@@ -199,7 +199,7 @@ class SV_WC_Payment_Gateway_Payment_Tokens_Handler {
 	 * @param int $user_id user identifier
 	 * @param SV_WC_Payment_Gateway_Payment_Token $token the token
 	 * @param string|null $environment_id optional environment id, defaults to plugin current environment
-	 * @return bool
+	 * @return int
 	 */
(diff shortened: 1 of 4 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-tokens-handler.php
```diff
@@ -363,33 +378,42 @@ public function delete_token( $user_id, SV_WC_Payment_Gateway_Payment_Token $tok
 			$environment_id = $this->get_environment_id();
 		}
 
-		// get existing tokens
-		$tokens = $this->get_tokens( $user_id, array( 'environment_id' => $environment_id ) );
+		// always clear the transient
+		$this->clear_transient( $user_id );
```
wvega: I think it would be good to clear the transient on `add_token()` and `update_token()` as well. The transient is short lived, but is still possible to get an outdated copy of the tokens in the next request.

PR #385, opened 2019-12-24
v5.5.2

unfulvio:
# Summary

This PR proposes a fix-workaround for a problem that has plagued some framework updates for mostly gateway plugins calling the API handler from the main plugin file to produce a notice upon plugin upgrade.

### Related Stories

- [CH 14277](https://app.clubhouse.io/skyverge/story/14277/updating-a-plugin-can-cause-temporary-fatal-errors)

## Details

The workaround consists of moving the API base methods necessary to produce a notice to the main plugin file, eliminating the possibility to load a file of the framework referencing a namespace no longer in use during a framework update.

As noted in [this comment]() unfortunately the benefits of such change will only be delivered on future plugin updates and customers may still experience the issue when they update a plugin using this release candidate. Furthermore, this PR does not address the root cause of the issue, which is likely due to caching (more explanations in the attached [CH story](https://app.clubhouse.io/skyverge/story/14277/updating-a-plugin-can-cause-temporary-fatal-errors)).

## UI changes

None.

## Implementation in plugins

- [ ] Look for `SV_WC_API_Base::require_tls_1_2()` references and replace usages to point to `SV_WC_Plugin::require_tls_1_2()`
- [ ] Look for `SV_WC_API_Base::is_tls_1_2_available()` references __in main plugin file only__ (important)* and replace usages to point to `SV_WC_Plugin::is_tls_1_2_available()`
- [ ] Plugins currently overriding `SV_WC_API_Base::require_tls_1_2()` should override `SV_WC_Plugin::require_tls_1_2()` instead.

\* - The reason why this is important is that `SV_WC_Plugin::is_tls_1_2_available()` does not pass a filter because there is no context for the `SV_WC_API_Base` that is used in the filter args. This is available in `SV_WC_API_Base` so plugins can still use `SV_WC_API_Base::is_tls_1_2_available()` _outside the main plugin file_ (which triggers the issue we are trying to fix).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "skyverge/wc-plugin-framework",
     "description": "The official SkyVerge WooCommerce plugin framework",
-    "version": "5.5.1",
+    "version": "5.5.2",
     "require-dev": {
         "lucatume/wp-browser": "^2.1"
     }
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "skyverge/wc-plugin-framework",
     "description": "The official SkyVerge WooCommerce plugin framework",
-    "version": "5.5.2",
+    "version": "5.5.1",
     "require-dev": {
         "lucatume/wp-browser": "^2.1"
     }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wc-plugin-framework",
-  "version": "5.5.2",
+  "version": "5.5.1",
   "title": "WooCommerce Plugin Framework",
   "author": "SkyVerge Team",
   "homepage": "https://github.com/skyverge/wc-plugin-framework#readme",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wc-plugin-framework",
-  "version": "5.5.1",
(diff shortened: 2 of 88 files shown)


ChaseWiseman:
Will make sure and give this a go this week, and we can include it in the next framework update for the gateways.

wvega:
@ChaseWiseman @unfulvio I propose to release https://github.com/skyverge/wc-plugin-framework/pull/386 as `v5.5.2` and bump the version for the modifications here.

That would allow a quick update for Authorize.Net to use the modifications from that PR and have a patched version ready for https://secure.helpscout.net/conversation/1046647489/119324?folderId=1528897 when WC 3.9 goes out.

ChaseWiseman:
@wvega @unfulvio makes sense to me!

This is a good case for keeping story changes separate from the FW versioning. The versioning can be done last, when tagging a new release. That way small fixes that we want to release sooner that something else aren't reliant on other PRs.

@unfulvio mind removing the versioning from this PR?

unfulvio:
@ChaseWiseman done and renamed this branch into `workaround-fw-upgrade-error` to avoid confusion (was previously named `v5.5.2`. I did the versioning and FW namespace update because I thought this was an urgent one that we would have pushed sooner.

PR #387, opened 2020-01-09
Workaround for occasional error on framework upgrade routines

unfulvio:
# Summary

This PR proposes a fix-workaround for a problem that has plagued some framework updates for mostly gateway plugins calling the API handler from the main plugin file to produce a notice upon plugin upgrade.

### Related Stories

- [CH 14277](https://app.clubhouse.io/skyverge/story/14277/updating-a-plugin-can-cause-temporary-fatal-errors)

## Details

The workaround consists of moving the API base methods necessary to produce a notice to the main plugin file, eliminating the possibility to load a file of the framework referencing a namespace no longer in use during a framework update.

As noted in [this comment]() unfortunately the benefits of such change will only be delivered on future plugin updates and customers may still experience the issue when they update a plugin using this release candidate. Furthermore, this PR does not address the root cause of the issue, which is likely due to caching (more explanations in the attached [CH story](https://app.clubhouse.io/skyverge/story/14277/updating-a-plugin-can-cause-temporary-fatal-errors)).

## UI changes

None.

## Implementation in plugins

- [ ] Look for `SV_WC_API_Base::require_tls_1_2()` references and replace usages to point to `SV_WC_Plugin::require_tls_1_2()`
- [ ] Look for `SV_WC_API_Base::is_tls_1_2_available()` references __in main plugin file only__ (important)* and replace usages to point to `SV_WC_Plugin::is_tls_1_2_available()`
- [ ] Plugins currently overriding `SV_WC_API_Base::require_tls_1_2()` should override `SV_WC_Plugin::require_tls_1_2()` instead.

\* - The reason why this is important is that `SV_WC_Plugin::is_tls_1_2_available()` does not pass a filter because there is no context for the `SV_WC_API_Base` that is used in the filter args. This is available in `SV_WC_API_Base` so plugins can still use `SV_WC_API_Base::is_tls_1_2_available()` _outside the main plugin file_ (which triggers the issue we are trying to fix).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "skyverge/wc-plugin-framework",
     "description": "The official SkyVerge WooCommerce plugin framework",
-    "version": "5.5.1",
+    "version": "5.5.2",
     "require-dev": {
         "lucatume/wp-browser": "^2.1"
     }
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "skyverge/wc-plugin-framework",
     "description": "The official SkyVerge WooCommerce plugin framework",
-    "version": "5.5.2",
+    "version": "5.5.1",
     "require-dev": {
         "lucatume/wp-browser": "^2.1"
     }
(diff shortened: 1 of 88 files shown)


unfulvio:
was previously #385

PR #397, opened 2020-02-18
Add token ID to core methods list item data

deivamagalhaes:
# Summary

This PR adds the token ID to the token data returned by `wc_get_account_saved_payment_methods_list` and used to populate the table.

## Details

I've used `get_token` instead of `get_id` because that is what we are using as the array key (https://skyverge.slack.com/archives/CTHBHPAUS/p1581988345009300).

## UI Changes

None.

## QA

No functionality to test, just code review.

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
@@ -190,7 +190,7 @@ class SV_WC_Payment_Gateway_My_Payment_Methods {
 	 */
 	public function add_payment_methods_list_item_id( $item, $token ) {
 
-		$item['id'] = $token->get_id();
+		$item['id'] = $token->get_token();
 
 		return $item;
 	}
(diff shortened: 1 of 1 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -173,6 +175,27 @@ protected function load_tokens() {
...
+	 *
+	 * @since 5.6.0-dev
+	 *
+	 * @param array $item individual list item from woocommerce_saved_payment_methods_list
+	 * @param \WC_Payment_Token $token payment token associated with this method entry
+	 * @return array
+	 */
+	public function add_payment_methods_list_item_id( $item, $token ) {
+
+		$item['id'] = $token->get_token();
```
wvega: Since this is a list of core tokens and the value represents what they call a token (instead of the ID), would it make sense to rename the entry to `token` as well? 

```suggestion
		$item['token'] = $token->get_token();
```

deivamagalhaes: I am not sure, because it is what _we_ call an `id`... Adding @ChaseWiseman for another opinion!

deivamagalhaes: I'll merge this as-is meanwhile to avoid nested PRs.

ChaseWiseman: @deivamagalhaes @wvega yeah I think naming it `token` makes sense, since at this point we're only dealing with a core token and not our wrapper of a token. This added value is also going to be available to everyone else, so it could be confusing if we add it as `id`

PR #412, opened 2020-02-20
Remove Framework table

wvega:
# Summary

This PR removes calls to methods that are no longer necessary now that framework tokens are included in WooCommerce's table of payment methods.

## Details

- I didn't remove leftover CSS and JS because we are still making small modifications here and there. 
- Tokens are no longer loaded on init with `load_tokens()`
- The framework's table is no longer rendered (via `render()`) at the bottom of the table
- `handle_payment_method_actions()` is no longer used because actions are now handled in core
- `render()` and `handle_payment_method_actions()` are now deprecated
- The following actions are now mapped to similar core actions for backwards compatibility
    - wc_{plugin_id}_before_my_payment_method_table
    - wc_{plugin_id}_after_my_payment_method_table
    - wc_payment_gateway_{gateway_id}_payment_method_deleted
- Restored support for custom payment methods actions via the  `wc_{plugin_id}_my_payment_methods_table_method_actions` and `wc_{plugin_id}_my_payment_methods_action_{action}` actions
- Deprecated several methods that render HTML for the framework table of payment methods

### Setup

- Configure NETBilling to use `dev-ch30010-remove-framework-table` version of the plugin framework (this branch)
- Have at least one saved payment method from the framework

#### Check the table is displayed correclty and JavaScript behavior is working

1. Navigate to My Payment Methods
    - [x] The Title, Details, and Default columns show appropriate values for the payment method
1. Click the Edit button
    - [x] The "Title" column switches to a text input
    - [x] The "Edit" button switches to the "Cancel" link
1. Enter a nickname for the payment method
1. Click Save
    - [x] The "Title" column switches back to plain text and displays the new nickname
    - [x] The "Cancel" link switches back to the "Edit" button

#### Test mapped actions

Add the following snippet to your site:

```php
add_action( 'wc_netbilling_before_my_payment_method_table', function() {

    echo '<div class="woocommerce-message">BEFORE</div>';
} );


add_action( 'wc_netbilling_after_my_payment_method_table', function() {

    echo '<div class="woocommerce-message">AFTER</div>';
} );


add_action( 'wc_payment_gateway_netbilling_payment_method_deleted', function() {

    wc_add_notice( 'wc_payment_gateway_netbilling_payment_method_deleted action was fired!' );
} );
```

1. Navigate to My Payments Methods
    - [x] A notice that reads "BEFORE" is visible before the table
    - [x] A notice that reads "AFTER" is visible after the table
1. Delete a framework saved payment method
    - [x] A notice with "wc_payment_gateway_netbilling_payment_method_deleted action was fired!" is shown when the page reloads

### Test custom payment methods actions

Add the following snippet to your site:

```php
add_action( 'wc_netbilling_my_payment_methods_table_method_actions', function( $actions, $token, $instance ) {

    $plugin_slug = $instance->get_plugin()->get_id_dasherized();

    $url = add_query_arg( [
        "wc-{$plugin_slug}-token"  => $token->get_id(),
        "wc-{$plugin_slug}-action" => 'custom',
    ] );

    $actions['custom'] = [
        'name' => 'Custom',
        'url'  => wp_nonce_url( $url, "wc-{$plugin_slug}-token-action" ),
    ];

    return $actions;
}, 10, 3 );

add_action( 'wc_netbilling_my_payment_methods_action_custom', function( $instance ) {

    wc_add_notice( 'wc_netbilling_my_parent_methods_action_custom action was fired!' );
} );
```

1. Navigate to My Payments Methods
    - [x] NETBilling payment methods have a Custom action button
1. Click the Custom action button
    - [x] A notice with "wc_netbilling_my_parent_methods_action_custom action was fired!" is shown when the page reloads


woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -103,13 +102,12 @@ public function init() {
...
-		add_action( 'woocommerce_after_account_payment_methods', array( $this, 'render' ) );
-		add_action( 'woocommerce_after_account_payment_methods', array( $this, 'render_js' ) );
+		// map Framework payment methods actions to WooCommerce actions for backwards compatibility
+		add_action( 'woocommerce_before_account_payment_methods', [ $this, 'before_payment_methods_table' ] );
+		add_action( 'woocommerce_after_account_payment_methods',  [ $this, 'after_payment_methods_table'] );
 
-		// handle payment method deletion, etc.
-		$this->handle_payment_method_actions();
+		// render JavaScript used in the My Payment Methods section
+		add_action( 'woocommerce_after_account_payment_methods', array( $this, 'render_js' ) );
```
deivamagalhaes: Can we switch to shorthand syntax?

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -423,46 +437,95 @@ public function render() {
...
+			'id'              => $this->get_plugin()->get_id(),
+			'slug'            => $this->get_plugin()->get_id_dasherized(),
+			'has_core_tokens' => (bool) wc_get_customer_saved_methods_list( get_current_user_id() ),
+			'ajax_url'        => admin_url( 'admin-ajax.php' ),
+			'ajax_nonce'      => wp_create_nonce( 'wc_' . $this->get_plugin()->get_id() . '_save_payment_method' ),
+			'i18n'            => array(
+				'edit_button'   => esc_html__( 'Edit', 'woocommerce-plugin-framework' ),
+				'cancel_button' => esc_html__( 'Cancel', 'woocommerce-plugin-framework' ),
+				'save_error'    => esc_html__( 'Oops, there was an error updating your payment method. Please try again.', 'woocommerce-plugin-framework' ),
+				'delete_ays'    => esc_html__( 'Are you sure you want to delete this payment method?', 'woocommerce-plugin-framework' ),
```
deivamagalhaes: Are we still using this one?

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -1147,86 +1210,16 @@ protected function save_token_data( SV_WC_Payment_Gateway_Payment_Token $token,
...
-
-					/**
-					 * My Payment Methods Custom Action.
-					 *
-					 * Fired when a custom action is requested for a payment method (e.g. other than delete/make default)
-					 *
-					 * @since 4.0.0
-					 * @param \SV_WC_Payment_Gateway_My_Payment_Methods $this instance
-					 */
-					do_action( 'wc_' . $this->get_plugin()->get_id() . '_my_payment_methods_action_' . sanitize_title( $action ), $this );
```
deivamagalhaes: Are we okay with removing support to custom actions here?

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
deivamagalhaes: We are, never mind.

deivamagalhaes:
I'm seeing the delete confirmation dialog twice... Maybe we don't need to add it because core already does?

deivamagalhaes:
Also, we should probably remove/deprecate the following unused methods:
- `get_table_html`
- `get_table_head_html`
- `get_table_body_html`
- `get_table_body_row_html`
- `get_table_body_row_data`
- `get_payment_method_actions_html`
- `get_payment_method_expiry_html`

deivamagalhaes:
> I'm seeing the delete confirmation dialog twice... Maybe we don't need to add it because core already does?

I just realized that is because I have 2 FW gateways enabled. It is possible this problem wasn't happening before because the JS/CSS selectors were gateway specific. We may have other similar issues. Each gateway should only handle it's own methods' delete button.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
wvega: @deivamagalhaes I think we are not. I worked on this prior to the discussion we had about removing/deprecating filters and didn't update it with the results of that discussion.

I believe we should update the method to remove the part that handles the delete action and handle custom actions only, for backwards compatibility purposes.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
wvega: Yes, fixed in 8dca0cc.

deivamagalhaes:
While working on the JS issue, I noticed another problem. When we removed `load_tokens` and replaced it `$this->get_plugin()->get_gateway()->get_payment_tokens_handler()->get_token`, we did not consider plugins with multiple gateways. I have 2 NETbilling saved methods, one credit card and one e-check. The e-check is not handled as a FW token/method by the table because `get_token_by_id` only returns tokens for the first gateway on the plugin (the credit card gateway).

I'll go ahead and update that method to handle multiple gateways.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -323,13 +317,38 @@ public function add_payment_methods_columns( $columns  = [] ) {
...
 	 * @since 5.6.0-dev
 	 *
+	 * @param string $token_id token string
+	 * @return SV_WC_Payment_Gateway_Payment_Token|null
+	 */
+	private function get_token_by_id( $token_id ) {
+
+		$token = null;
+
+		foreach ( $this->get_plugin()->get_gateways() as $gateway ) {
```
wvega: @deivamagalhaes good catch! Thank you for updating the method.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
wvega: @deivamagalhaes I made a couple of related updates:

* `handle_payment_method_actions()` is no longer deprecated and handle custom actions only now: 2852e99
* `wc_{plugin_id}_my_payment_methods_table_method_actions` is triggered when we add the Edit and Save buttons to the list of core actions: 032d459

wvega:
@deivamagalhaes I deprecated most of the methods you noted in a previous comment. I kept `get_table_body_row_data()` for now, because we may be able to keep support for the `wc_{plugin_id}_my_payment_methods_table_body_row_data` filter it contains. So we will either modify the method or deprecate it in another story. I have to take a closer look.

I also added another test case to the QA steps to check support for custom actions.

Could you please take another look at this PR when you get a chance?

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -795,21 +779,6 @@ protected function get_table_body_row_data( $token ) {
...
-	 *
-	 * @param SV_WC_Payment_Gateway_Payment_Token $token token object
-	 * @return string
-	 */
-	protected function get_payment_method_title( $token ) {
-
-		return $this->get_payment_method_title_html( $token );
-	}
-
-
```
deivamagalhaes: Out for curiosity, why did you remove this one instead of deprecating it? I noticed you deprecated at least one other protected method...

wvega: That one was deprecated in February 2018, but we didn't add a notice or a TODO, only the deprecated tag. Since it's been almost two years, I think is ok to remove it.

wvega:
>the row data filter (should we write a story for that so we don't forget)

Yes, that's a good idea. I'll write one today, with notes about the proposal and some of the discussion on Slack.

PR #413, opened 2020-02-20
Fix Default column for core tokens

deivamagalhaes:
# Summary

This PR fixes the Default column so it displays the "Default" mark for FW tokens and core tokens alike. 

## Details

Because the pre-existing filter expects a FW token as a param, it is only applied if the token is a FW token.

## UI Changes

- The Default column displays the mark for core tokens as well: https://cloud.skyver.ge/lluyzZpD

### Setup

- Have at least one FW payment method saved
- Have at least one core payment method saved (recommended gateway: Stripe)

### Steps

1. Navigate to Payment Methods
1. Make the core token the default, if needed
    - [x] The "Default" column for the core token shows the Default mark
    - [x] The "Default" column for the FW token does not show the Default mark
1. Make the FW token the default
    - [x] The "Default" column for the core token does not show the Default mark
    - [x] The "Default" column for the FW token shows the Default mark

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
@@ -379,10 +379,7 @@ class SV_WC_Payment_Gateway_My_Payment_Methods {
 	 */
 	public function add_payment_method_default( $method ) {
 
-		if ( $token = $this->get_token_by_id( $method ) ) {
-
-			echo $this->get_payment_method_default_html( $token );
-		}
+		echo $this->get_payment_method_default_html( ! empty( $method['is_default'] ), $this->get_token_by_id( $method ) );
 	}
 
 
@@ -765,7 +762,7 @@ class SV_WC_Payment_Gateway_My_Payment_Methods {
 
 		$method = array(
 			'title'   => $this->get_payment_method_title_html( $token ),
-			'default' => $this->get_payment_method_default_html( $token ),
+			'default' => $this->get_payment_method_default_html( $token->is_default(), $token ),
 			'details' => $this->get_payment_method_details_html( $token ),
 			'actions' => $this->get_payment_method_actions_html( $token ),
 		);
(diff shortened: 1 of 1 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -857,22 +854,28 @@ protected function get_payment_method_title_html( SV_WC_Payment_Gateway_Payment_
 	 *
 	 * @since 5.1.0
 	 *
-	 * @param SV_WC_Payment_Gateway_Payment_Token $token token object
+	 * @param boolean $is_default true if the token is the default token
+	 * @param SV_WC_Payment_Gateway_Payment_Token|null $token FW token object, only set if the token is a FW token
 	 * @return string
 	 */
-	protected function get_payment_method_default_html( SV_WC_Payment_Gateway_Payment_Token $token ) {
+	protected function get_payment_method_default_html( $is_default, SV_WC_Payment_Gateway_Payment_Token $token = null ) {
```
asseti6: @deivamagalhaes assuming our min WP/WC PHP version does not support type casting here?  If it does this bool really should be type casted and set to a default of false.

deivamagalhaes: We support PHP 5.6, so yes, no scalar type declarations for us. I can still set the default though, if you think that would be helpful.

asseti6: Yeah I think we should default to that on a bool parameter.  Can easily have someone do a strict comparison later and be unaware.

asseti6: Ah my spelling sucks.. lets try this again: Yeah I think we should default to that on a bool parameter for readability. And it saves a step keeping things clear in the future where we do upgrade PHP versions.

PR #417, opened 2020-02-22
Improve appearance of payments table

wvega:
# Summary

This PR updates the CSS rules for the payment methods table to apply to the core table of payment methods in order to make the table look better with the new columns and fit the available space.

## UI Changes

The table before these modifications:

Desktop: https://cloud.skyver.ge/mXuq7RZD
Desktop (editing): https://cloud.skyver.ge/p9uKzdno
Mobile: https://cloud.skyver.ge/d5u0bl8o

The table after these modifications:

Desktop: https://cloud.skyver.ge/12u1E9BG
Desktop (editing): https://cloud.skyver.ge/xQugvdep
Mobile: https://cloud.skyver.ge/nOuNWEB8

### Setup

- Have at least two framework tokens
- Have at least one core token

### Steps

1. Navigate to Payment Methods
    - [x] the table width is closer to the available width
    - [x] content on each row is vertically aligned to the middle of the row
    - [x] Action buttons are aligned to the right
1. Click the Edit button for one of the payment methods
    - [x] The width of the table doesn't change
    - [x] The width of the nickname input field is constrained to the width of the column
    - [x] Other rows appear disabled (opacity is less than 1)


wvega:
@deivamagalhaes I changed the alignment of action buttons to `right` on this PR because that's they way they showed in the framework's version of the table. However, I noticed you had set it to `center` recently.

Should we use `center` instead? I can revert the commit easily (4ae9df2) if necessary.

Either `right` or `center` look good to me on the mobile version, but I think `right` looks better on wider screens. What do you think?

deivamagalhaes:
I'm still not sure about the buttons right-aligned, I think it makes the size difference more visible. Maybe it's something with the margins, I'm not sure.

Also, when editing and there are more buttons, there is no spacing between the "Save" and the next button: https://cloud.skyver.ge/qGuozALq

When editing on mobile, should the input be this large? https://cloud.skyver.ge/Qwu7QxZY

wvega:
>On mobile, I think the buttons centered looks a lot better, and if I'm not mistaken they were already centered on the mobile version of the FW table.

I saw them right aligned on a copy of Authorize.Net I'm running on a different site, but I'm starting to see the issues with having them right aligned on the new table.

>Maybe it's something with the margins, I'm not sure.

I noticed they are not perfectly aligned to the right because of the `&nbsp;` characters that are printed between the `<a>` elements and end up shifting the buttons a bit. It happens when the buttons are centered too, but is less noticeable. I'll revert the commit.

>Also, when editing and there are more buttons, there is no spacing between the "Save" and the next button: https://cloud.skyver.ge/qGuozALq

That seems to occur for both `right` and `center` alignment as well. We are setting `display: inline` when we make the button visible, but all the other buttons use `inline-block`. It looks a bit better with with `inline-block`: https://cloud.skyver.ge/2NuBAkbn

>When editing on mobile, should the input be this large? https://cloud.skyver.ge/Qwu7QxZY

I think so, that's the size it had on the FW table: https://cloud.skyver.ge/JruWwlOv

wvega:
@deivamagalhaes I forgot to request another review last time I updated this one. Requesting again now.

Would you please take a look at my last comment and modifications?

wvega:
>One thing I noticed is that the custom actions button (defined using the filter) looks a bit weird (goes to another line): https://cloud.skyver.ge/12u1D627

@deivamagalhaes the button goes to the next line because there is not enough room for all the buttons in a single line: https://cloud.skyver.ge/L1ugk4xv. However, even without the padding, the cell seems too small for all the buttons to fit on that screen size, because it's limited by the size of the table and the content area.

I think it would be hard to fix for everyone. We could change the font-size, the margin, or the button's size to make them fit, but I think it would work for this specific case only. If there are more buttons or the theme uses a very different appearance it would likely start looking weird again.

What do you think?

deivamagalhaes:
I thought it wasn't because of space because "Edit", "Delete" and "Make default" fit in one line, and "Custom" and "Delete" have the same number of letters, but you're right, it's the same number of letters but it's wider. So I suppose it's fine! 🤷‍♀

PR #418, opened 2020-02-24
Deprecate get my payment methods table body row method

unfulvio:
### Summary

Deprecates `get_table_body_row_data()` and related filter. 

## Details

I deprecated this method like others in the same class. Some similar filters have been removed as well. Being a framework and not a user-facing plugin, perhaps we don't need to deprecate filters when it is not possible to map them to other hooks. Also, I don't think we should use the hook deprecator from the fw, but rather leave that choice to individual child implementations (plugins) that may choose to do so. 

### QA

- [ ]  The method `get_table_body_row_data()` triggers a deprecation warning when invoked
- [ ]  The filter `'wc_' . $this->get_plugin()->get_id() . '_my_payment_methods_table_body_row_data'` is no longer applicable


unfulvio:
As mentioned in our WC weekly I'm going to write a separate story to handle hooks that were removed/deprecated as in this PR.

PR #419, opened 2020-02-24
Handle deprecated filter hooks

unfulvio:
### Summary

Handles deprecated hooks that have no direct replacement.

### Additional details

Instead of duplicating this across all plugins, we can deprecate some hooks directly in the plugin. But there's likely no need to use the hooks deprecator: when the payment methods handler is initialized, it could look if third party code is trying to apply certain filters/actions, and then raise a deprecation notice accordingly.

### UI changes

None.

### QA

- [x] If you try to use any of the deprecated filters, it should trigger a notice.

### DIFF
--- a/woocommerce/changelog.txt
+++ b/woocommerce/changelog.txt
@@ -2,6 +2,7 @@
 
 2020.nn.nn - version 5.6.0-dev
  * Tweak - Migrate payment tokens to be compatible with WooCommerce core payment tokens
+ * Dev - Deprecate some filter hooks in the payment methods table
 
 2020.01.20 - version 5.5.4
  * Tweak - Add a link to the site's terms and conditions page below Apple Pay buttons when available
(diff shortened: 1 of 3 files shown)


unfulvio:
@deivamagalhaes uhm, then it would run the check on any plugin not just gateways? I am not sure, feels to me in that class it has immediate context where the hooks were related to. Also, doesn't the deprecator need to be initialized by the child plugin? If yes, what if they don't already?

deivamagalhaes:
I believe every plugin runs `init_hook_deprecator` and `get_deprecated_hooks` anyway, it is just empty by default. I don't know, it just feels a bit weird to implement brand new code for something the FW already does... 🤔

unfulvio:
@deivamagalhaes fair enough, I've implemented this when I was about to go to sleep, it didn't make much sense as I originally though. Should work with the deprecator then. See my latest changes. In the eventuality a plugin needs to define a different deprecation for the same filter, it can still do so via that `array_merge`: deprecations defined by child plugin should overwrite the same keys (I doubt this will happen though).

PR #421, opened 2020-02-25
Restore load_tokens() call upon initializing payment methods

unfulvio:
## Summary

Restores `load_tokens()` from the `init()` method in the My Payment Methods handler.

## Additional details

As @wvega pointed out, this helps initializing and maybe migrating any additional tokens before they reach WooCommerce internals to output the tokens table which we now filter too.

## QA

- [ ] If you have existing tokens which have not been migrated when opening the My Payments Method, these should be migrated now with the changes proposed in PR (tested with Cybersource).

### DIFF
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -728,7 +728,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 			$this->token = null;
 
 			// TODO probably this exception should be logged to the gateway log {FN 2020-02-25}
-			$saved = 0;
+			$saved = $token->get_id();
 		}
 
 		return $saved;
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -592,7 +592,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 					'user_id'    => $this->get_user_id(),
 					'gateway_id' => $this->get_gateway_id(),
 				]
-			);;
+			);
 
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -713,10 +713,22 @@ public function save() {
...
 		}
 
-		$saved = $token->save();
+		$token->apply_changes();
 
-		if ( $saved ) {
-			$this->token = $token;
+		try {
+
+			$saved = $token->save();
```
unfulvio: while testing with Authorize.Net this threw an exception. Looks like it's because Authorize.Net uses its own token object so that needs some tweak. However, that made me realize that we should be able to catch any exception that would result in a token not being saved, regardless of where this occurrence happens. I would say that it should be logged, but there's no plugin/gateway context here...

wvega: >Looks like it's because Authorize.Net uses its own token object so that needs some tweak.

I think I saw something similar last week. In my case Authorize.Net needed to accept a core token in addition to an array of data, like the framework token does now.

However, after that, I found another issue that occurs because the credit card tokens that Authorize.Net API returns don't have an expiry date. `WC_Payment_Token_CC` has a validation method that requires the expiry year to be a four characters string and the expiry month a two characters string. Authorize.Net tokens do not currently pass that validation and cause an exception to be thrown from the the tokens data store.

>I would say that it should be logged, but there's no plugin/gateway context here...

We can probably catch the exceptions and log the error in the payment tokens handler. I believe we interact with tokens through the tokens handler in most places, so that should cover a good portion of the cases.

wvega: Hand't looked at the surrounding code when I responded. Still think that we could catch and log an exception in the tokens handler. For that, we could throw a framework exception in the `catch` block here.

However, I think that for the scope of this story, the TODO is enough.

PR #422, opened 2020-02-25
Catch an error when saving a token

unfulvio:
## Summary

When `WC_Data::save()` triggers an Exception, the framework token wrapper should catch it and log the message.

## Additional details

There is no access to the gateway handler and thus the gateway/plugin/logger from the token object context. Therefore I thought we could normalize exceptions thrown by WC into the framework one for gateways, and catch that from other places where the framework tries to save a token.

## Acceptance

- [ ] Exceptions shouldn't halt plugin/framework. 
- [ ] Exception messages should be logged.

### DIFF
--- a/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
+++ b/woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
@@ -676,6 +676,7 @@ class SV_WC_Payment_Gateway_Payment_Token {
 	 * @since 5.6.0-dev.1
 	 *
 	 * @return int ID of the token saved as returned by {@see \WC_Payment_Token::save()}
+	 * @throws SV_WC_Payment_Gateway_Exception when saving the parent token hits an error
 	 */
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -725,10 +726,7 @@ public function save() {
 
 		} catch ( \Exception $e ) {
 
-			$this->token = null;
-
-			// TODO probably this exception should be logged to the gateway log {FN 2020-02-25}
-			$saved = $token->get_id();
+			throw new SV_WC_Payment_Gateway_Exception( $e->getMessage() );
```
wvega:
```suggestion
			throw new SV_WC_Payment_Gateway_Exception( 'Could not save payment token. ' . $e->getMessage() );
```

The exception was logged as follows, so I thought it would be good to expand the message to indicate what we were trying to do:

    02-25-2020 @ 13:24:21 - Invalid or missing payment token fields.

PR #423, opened 2020-02-26
Namespace the payment methods JS handler class to avoid conflicts

unfulvio:
## Summary

On installations that use multiple frameworked gateways simultaneously, the same JS handlers in `window.SV_WC_Payment_Methods_Handler` might overwrite each other - therefore we could make these names unique to match the current version of the framework to avoid collisions.

## Additional details

I went for `SV_WC_Payment_Methods_Handler_v5_5_4` and matched this to the script handle used by WordPress to enqueue js and css files.

## QA

- [ ] Scripts and styles are enqueued properly

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
@@ -119,15 +119,15 @@ class SV_WC_Payment_Gateway_My_Payment_Methods {
 
 
 	/**
-	 * Enqueue frontend CSS/JS
+	 * Enqueues frontend scripts and styles.
 	 *
 	 * @since 4.0.0
 	 */
 	public function maybe_enqueue_styles_scripts() {
 
-		$handle = 'sv-wc-payment-gateway-my-payment-methods';
+		$handle  = 'sv-wc-payment-gateway-my-payment-methods-v5_5_4';
 
-		wp_enqueue_style( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/css/frontend/' . $handle . '.min.css', array( 'dashicons' ), SV_WC_Plugin::VERSION );
+		wp_enqueue_style( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/css/frontend/' . $handle . '.min.css', [ 'dashicons' ], SV_WC_Plugin::VERSION );
 
 		wp_enqueue_script( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/js/frontend/' . $handle . '.min.js', [ 'jquery' ], SV_WC_Plugin::VERSION );
 	}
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -119,15 +122,16 @@ public function init() {
...
 	 * @since 4.0.0
 	 */
 	public function maybe_enqueue_styles_scripts() {
 
-		$handle = 'sv-wc-payment-gateway-my-payment-methods';
+		$js_class = sprintf( '%s_v5_5_4', $this->js_handler_base_class_name );
+		$handle   = strtolower( str_replace( '_', '-', $js_class ) );
 
-		wp_enqueue_style( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/css/frontend/' . $handle . '.min.css', array( 'dashicons' ), SV_WC_Plugin::VERSION );
+		wp_enqueue_style( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/css/frontend/' . $handle . '.min.css', [ 'dashicons' ], SV_WC_Plugin::VERSION );
```
wvega:
```suggestion
		wp_enqueue_style( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/css/frontend/sv-wc-payment-gateway-my-payment-methods.min.css', [ 'dashicons' ], SV_WC_Plugin::VERSION );
```

We should keep the same file name.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -119,15 +122,16 @@ public function init() {
...
 
 		wp_enqueue_script( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/js/frontend/' . $handle . '.min.js', [ 'jquery' ], SV_WC_Plugin::VERSION );
```
wvega:
```suggestion
		wp_enqueue_script( $handle, $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/js/frontend/sv-wc-payment-gateway-my-payment-methods.min.js', [ 'jquery' ], SV_WC_Plugin::VERSION );
```

We should keep the same file name here as well.

unfulvio:
@wvega sorry I forgot to add that in the PR - actually locally I was testing with a modified file name, but I don't think we need that, I think we can keep it as it is now, no?

unfulvio:
merging this so I can run some tests from the same branch, if we disagree on the js file name we can fix it later :)

wvega:
>actually locally I was testing with a modified file name, but I don't think we need that, I think we can keep it as it is now, no?

@unfulvio yes, we can keep it as it is now. That was my intention too, but I think the wording in my review note was confusing. No fix needed :)

PR #424, opened 2020-02-26
Prevent uncaught error in User Profile page when Subscriptions is active

wvega:
# Summary

This PR disables the notice that Subscriptions shows when a token is set as default and there are other tokens from the save gateway that are associated to subscriptions, if the `wc_add_notice()` function is not defined.

Disabling the notice prevents an `Uncaught Error: Call to undefined function wc_add_notice()` error from being triggered in the User Profile page.

## QA

We should be able to test having at least two tokens of the same gateway for any of our gateways. I found that scenario was easier to reproduce using Authorize.Net, even considering the necessary modification to `WC_Authorize_Net_CIM_Payment_Profile`.

### Setup

- Update `WC_Authorize_Net_CIM_Payment_Profile::__construct` to wrap all but the last line inside `if ( is_array( $data ) ) { }`
- Configure Subscriptions and create a subscription product
- Configure Authorize.Net Credit Card gateway using the version of the framewrok in the epic branch (`v5.6.0.x-dev`) (without these modifications)

#### Reproduce the error

1. Purchase the subscription product using credit card number `4007000000027`
1. Purchase another product using credit card number `4012888818888` and save the payment method to your account
1. Go to the User Profile page of the user that placed the orders
    - [x] `Uncaught Error: Call to undefined function wc_add_notice()` is triggered while trying to render the Tokens Editor

#### Fix the error

1. Update Authorize.Net to use version `dev-ch30782-prevent-uncaught-error-in-user-profile-page` of the framework
1. Go to the User Profile page of the user that placed the orders
    - [x] No `Uncaught Error: Call to undefined function wc_add_notice()` is triggered
1. Change the expiration date or set the other token as default and save
    - [x] No `Uncaught Error: Call to undefined function wc_add_notice()` is triggered
    - [x] Modifications are saved successfully

### DIFF
--- a/woocommerce/payment-gateway/integrations/class-sv-wc-payment-gateway-integration-subscriptions.php
+++ b/woocommerce/payment-gateway/integrations/class-sv-wc-payment-gateway-integration-subscriptions.php
@@ -139,11 +139,13 @@ class SV_WC_Payment_Gateway_Integration_Subscriptions extends SV_WC_Payment_Gate
 
 
 	/**
-	 * Disables Subscription's default payment token change notice if the current request is not a frontend request.
+	 * Disables Subscription's default payment token change notice if wc_add_notice() is not defined
+	 *
+	 * This prevents an uncaught error from being triggered when tokens are retrieved and saved in the user profile page.
 	 *
 	 * @internal
 	 *
-	 * @see \WooCommerce::is_request()
+	 * @see \WCS_My_Account_Payment_Methods::display_default_payment_token_change_notice()
 	 *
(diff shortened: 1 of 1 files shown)


deivamagalhaes:
Oh, it's Subscriptions' action, not ours! Got it! 👍

wvega:
>I am still working on reproducing this one (I confess I am trying with another gateway).

I think it's easier to trigger if each token is associated to a subscription, that way no matter which one you set as default, there is another that has a subscription and triggers the notice.

PR #429, opened 2020-02-28
Delete legacy tokens and remote tokens when an associated core token is deleted

wvega:
# Summary

This PR removes local and remote token data when a core token is deleted directly. For example, when a user deletes a payment method from the Payment Methods page.

## Details

Also, as a result of the changes in this PR, `payment_method_deleted()` was no longer able to retrieve a framework token object and couldn't determine whether the deleted token was a framework token or not.

This PR updates that method to use information available in core token object to decide whether the 'wc_payment_gateway_{gateway_id}_payment_method_deleted' filter should be triggered or not.

### Setup

- Configure master version of Authorize.Net in test mode

### Steps

1. Add  credit card `4012888818888` as a saved payment method
1. Confirm that the token was stored as user meta:
    `wp user meta get {user_id} _wc_authorize_net_cim_credit_card_payment_tokens_test --format=json`
1. Confirm that a the tokens transient for the credit card gateway includes the token:
    `wp transient list --search="wc_sv_tokens*"`
1. Make sure tokens will be migrated when you install the next version of Authorize.Net:
    `wp user meta delete {user_id} _wc_authorize_net_cim_credit_card_payment_tokens_test_migrated`

Switch to branch `authorize.net-switch-to-core-tokens` and update `composer.json` to use version `dev-ch31230/delete-legacy-and-remote-token-data` (this branch) of the framework (run composer update)

1. Go to the Payment Methods page
1. Delete credit card ending in `8888`
1. Confirm that the token is no longer stored in user meta:
    `wp user meta get {user_id} _wc_authorize_net_cim_credit_card_payment_tokens_test --format=json`
    - [x] The token is not stored in user meta
1. Confirm that a the tokens transient for the credit card gateway **does not** include the token:
    `wp transient list --search="wc_sv_tokens*"`
    - [x] The token is not included
1. Delete the transient or wait until it expires to force the framework to load tokens again
    - [x] The deleted token is not created again

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
@@ -494,31 +494,37 @@ class SV_WC_Payment_Gateway_My_Payment_Methods {
 	 *
 	 * @since 5.6.0-dev
 	 *
-	 * @param int $token_id the ID of a core token
+	 * @param int $core_token_id the ID of a core token
 	 * @param \WC_Payment_Token $core_token the core token object
 	 */
-	public function payment_token_deleted( $token_id, $core_token ) {
+	public function payment_token_deleted( $core_token_id, $core_token ) {
 
-		$token = $core_token instanceof \WC_Payment_Token ? $this->get_token_by_id( $core_token->get_token() ) : null;
+		$token_id = null;
 
-		if ( $token instanceof SV_WC_Payment_Gateway_Payment_Token ) {
+		// find out if the core token belongs to one of the gateways from this plugin
+		foreach ( $this->get_plugin()->get_gateways() as $gateway ) {
 
-			// confirm the token was deleted from the Payment Methods screen
-			if ( (int) $token_id === (int) get_query_var( 'delete-payment-method' ) ) {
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-my-payment-methods.php
```diff
@@ -494,32 +494,38 @@ public function after_payment_methods_table( $has_methods ) {
...
 	 * @param \WC_Payment_Token $core_token the core token object
 	 */
-	public function payment_token_deleted( $token_id, $core_token ) {
+	public function payment_token_deleted( $core_token_id, $core_token ) {
 
-		$token = $core_token instanceof \WC_Payment_Token ? $this->get_token_by_id( $core_token->get_token() ) : null;
+		$token_id = null;
 
-		if ( $token instanceof SV_WC_Payment_Gateway_Payment_Token ) {
+		// find out if the core token belongs to one of the gateways from this plugin
```
deivamagalhaes: Maybe it would be good to add a comment saying we can't use `get_token_by_id` in this scenario because the FW token was already deleted, for future reference.

wvega: Good idea. I added another comment in 024f2dc

PR #431, opened 2020-03-04
Add admin notes helper

ChaseWiseman:
# Summary

Add some base helpers for dealing with WooCommerce Admin notes.

## Details

Adds a base helper class for retrieving admin notes. This can be enhanced in the future with much more functionality as we define our approach to dealing with admin notes.

As an example for other plugins, we've implemented the gateway "misconfiguration" notice using these methods.

## UI Changes

The notice text doesn't change, but now appears in WC Admin: https://cloud.skyver.ge/04uKNZqD

## QA

- Ensure you're running WooCommerce 4.0
- Point Authorize.Net's composer to `dev-ch31505/admin-notes-helper`
- Clear all `authorize_net` options from your DB

1. Activate
    - [x] No error notice is shown
1. Go to the Credit Card gateway settings page
1. Enable the gateway, but don't fill in any settings
1. Save
    - [x] You should now see the enhanced admin note error
1. Reload the page
    - [x] The same error is displayed, but only once
1. Navigate to the Plugins page
    - [x] The same error is display, in classic form
1. Navigate to the Orders page
    - [x] The enhanced note is displayed with Configure and Dismiss buttons
1. Click Configured
    - [x] You're taking to the gateway settings page
1. Navigate back to the Orders page
1. Click dismiss
1. Reload the page
    - [x] The note is no longer displayed
1. Go to the Credit Card gateway settings page
    - [x] The note is displayed again
1. Expand the "W" notices tab at the top right of the enhanced UI
    - [x] The same classic version of the error notice is _not_ present

## Before merge

### DIFF
--- a/woocommerce/admin/Notes_Helper.php
+++ b/woocommerce/admin/Notes_Helper.php
@@ -94,7 +94,6 @@ class Notes_Helper {
 			/** @var WooCommerce_Admin_Notes\DataStore $data_store */
 			$data_store = \WC_Data_Store::load( 'admin-note' );
 
-			// We already have this note? Then exit, we're done.
 			$note_ids = $data_store->get_notes_with_name( $name );
 
 		} catch ( \Exception $exception ) {}
--- a/woocommerce/admin/Notes_Helper.php
+++ b/woocommerce/admin/Notes_Helper.php
@@ -1,0 +1,106 @@
+<?php
+/**
+ * WooCommerce Plugin Framework
+ *
+ * This source file is subject to the GNU General Public License v3.0
+ * that is bundled with this package in the file license.txt.
+ * It is also available through the world-wide-web at this URL:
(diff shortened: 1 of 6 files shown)


woocommerce/admin/Notes_Helper.php
```diff
@@ -0,0 +1,106 @@
...
+	public static function get_note_ids_with_name( $name ) {
+
+		$note_ids = [];
+
+		try {
+
+			/** @var WooCommerce_Admin_Notes\DataStore $data_store */
+			$data_store = \WC_Data_Store::load( 'admin-note' );
+
+			// We already have this note? Then exit, we're done.
```
wvega:
```suggestion
```

I didn't understand this comment. Maybe is not needed here?

ChaseWiseman: whoops, yep, copy pasta!

PR #433, opened 2020-03-05
Fix changing the ID of a token using the Tokens Editor creates a copy of the token

wvega:
# Summary

This PR updates the Token Editor to update existing tokens without creating new token records on the database.

## Details

Using build_token() to create tokens with the modified data works in most cases, because the Framework token normally updates the associated core token record instead of creating a new one. However, if the ID of the the Framework token changes, it can no longer find the associated core token record and ends up creating a new one in the database.

This PR uses build_token() to store data for new tokens only and uses setter methods to apply modifications to the Framework token object for existing ones.

### Setup

- Configure CyberSource from `cybersource-switch-to-core-tokens` branch
- Update `composer.json` to use version `dev-ch31591/prevent-duplicate-tokens-on-tokens-editor` of the framework (this branch)
- Save a payment method

### Steps

1. Go to the User Profile page for the user that saved the payment method
    - [x] The token editor includes the saved payment
1. Change the Token ID of the payment method and click Save
    - [x] The token editor includes the saved payment with the new ID
    - [x] No extra saved payment methods were added to the table
1. Change the Card Type, Last Four, Expiration, and Default values for the saved payment method
1. Click Save
    - [x] The token editor includes the saved payment with the modified values
    - [x] No extra saved payment methods were added to the table
1. Click Add New to add a new saved payment method and enter values for all the fields
1. Click Save
    - [x] The new payment method was saved correctly

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
+++ b/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
@@ -169,8 +169,7 @@ class SV_WC_Payment_Gateway_Admin_Payment_Token_Editor {
 			$token_id          = $data['id'];
 			$original_token_id = isset( $data['original_id'] ) ? $data['original_id'] : '';
 
-			unset( $data['id'] );
-			unset( $data['original_id'] );
+			unset( $data['id'], $data['original_id'] );
 
 			if ( ! $token_id ) {
 				continue;
--- a/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
+++ b/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
@@ -166,7 +166,7 @@ class SV_WC_Payment_Gateway_Admin_Payment_Token_Editor {
 
 		foreach ( $tokens as $data ) {
 
-			$token_id          = $data['id'];
+			$token_id          = isset( $data['id'] ) ? $data['id'] : '';
 			$original_token_id = isset( $data['original_id'] ) ? $data['original_id'] : '';
 
 			unset( $data['id'], $data['original_id'] );
--- a/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
+++ b/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
@@ -336,6 +336,7 @@ class SV_WC_Payment_Gateway_Admin_Payment_Token_Editor {
 	 * @param SV_WC_Payment_Gateway_Payment_Token the payment token object to update
 	 * @param string $token_id the token ID
 	 * @param array $data the token data
+	 * @return SV_WC_Payment_Gateway_Payment_Token
 	 */
 	protected function update_token( $token, $token_id, $data ) {
 
--- a/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
+++ b/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
@@ -186,6 +186,7 @@ class SV_WC_Payment_Gateway_Admin_Payment_Token_Editor {
 
 				$token = $original_token_id ? $this->get_gateway()->get_payment_tokens_handler()->get_token( $user_id, $original_token_id ) : null;
 
+				// update the token props if a token with the original ID already exists to avoid creating another core token in SV_WC_Payment_Gateway_Payment_Token::save()
 				if ( $token instanceof SV_WC_Payment_Gateway_Payment_Token ) {
 					$built_tokens[ $token_id ] = $this->set_token_props( $token, $token_id, $data );
 				} else {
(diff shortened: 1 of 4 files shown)


woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
```diff
@@ -721,8 +734,6 @@ public function save() {
 			}
 		}
 
-		$token->apply_changes();
```
wvega: @unfulvio do you remember why we added `apply_changes()` in https://github.com/skyverge/wc-plugin-framework/commit/9848b634ea78917a7f73394080969827e3210ecf?

I found out that the token data store relies on pending changes to decide what information to update on the database. If we apply changes before calling save(), changed props are not persisted on update.

See [here](https://github.com/woocommerce/woocommerce/blob/e39db20eca4bbcf1b8b12c456393448f76ec8aa8/includes/data-stores/class-wc-payment-token-data-store.php#L95-L103).

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
```diff
@@ -318,6 +328,33 @@ protected function build_token( $user_id, $token_id, $data ) {
...
+
+			if ( is_callable( [ $token, "set_{$key}" ] ) ) {
+				$token->{"set_{$key}"}( $value );
+			}
+		}
+
+		$token->set_id( $token_id );
+
+		return $token;
+	}
```
unfulvio: I'm wondering if this could cause a bit of confusion, since the token is not saved, unlike with the `update_tokens()` method. Should it be an internal private method or part of the `save()` method for now? Or should it be part of the `build_token()`? I don't have a great alternative of doing this, just throwing a bit of feedback; I'm not too strongly opinionated in suggesting a different way here.

wvega: You make a good point about `update_token()` being misleading. I think is good to keep the method separate to avoid adding more complexity to `save()` or `build_token()`. Maybe we can rename it to better reflect what it does? What do you think about using `set_token_props()` or something similar?

I made the method `protected` because it's a companion for `build_token()` and that one was already `protected`. However, if we don't see sub-classes having to overwrite this one soon, we can keep it private for now.

woocommerce/payment-gateway/payment-tokens/class-sv-wc-payment-gateway-payment-token.php
wvega: No worries. When I reviewed that change it made sense to me as well. I wasn't aware of the way the data store uses those changes.

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
```diff
@@ -159,15 +159,18 @@ public function display_tokens( $user_id ) {
...
+		$tokens = ( isset( $_POST[ $this->get_input_name() ] ) ) ? $_POST[ $this->get_input_name() ] : [];
 
-		$built_tokens = array();
+		$default_token_id = SV_WC_Helper::get_posted_value( $this->get_input_name() . '_default' );
+		$built_tokens     = [];
 
 		foreach ( $tokens as $data ) {
 
-			$token_id = $data['id'];
+			$token_id          = $data['id'];
```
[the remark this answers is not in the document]
wvega: I don't understand this one :P 

Is that about checking that `$data['id']` is set before trying to access it? If that's the case, I think that's a good idea.
```suggestion
			$token_id          = isset( $data['id'] ) ? $data['id'] : '';
```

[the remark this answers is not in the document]
unfulvio: sorry I accidentally wrote "hard" I meant "harden" of course ahah

and yes that's what I was suggesting in your commit suggestion above

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
unfulvio: Maybe something like apply token changes or reinitialize token or (re) set token props as you say. If the method stays `private` then it's a bit less of a concern as we can come up with a better idea later if needed. If you stick to `private` can make a note to reconsider naming if opening to public/making protected. Thanks!

ChaseWiseman:
@wvega is there anything else needed for this PR?

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
wvega: @unfulvio I changed the name to `set_token_props()`, the method still does not apply the changes as in `WC_Data::apply_changes()`, so set token props sounded better to me. I was also going to make it private, but found out that we would need to overwrite it in Authorize.Net to also [update the `customer_profile_id` and `billing` data](https://github.com/skyverge/wc-plugins/blob/d3abef59de5d26167f1e06814df8668b80703658/woocommerce-gateway-authorize-net-cim/includes/class-wc-authorize-net-cim-payment-profile-editor.php#L49-L61).

wvega:
@ChaseWiseman yes, I had to address one last bit of feedback regarding the `set_token_props()` method but didn't dedicate time to doing that. I finally pushed an update today and will try to coordinate with Fulvio to get it merged soon.

woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-payment-token-editor.php
unfulvio: oh that sounds good @wvega thanks for getting back to this after many days, I know it's not ideal, we should get this merged then

unfulvio:
seems we are ready to merge this one

PR #444, opened 2020-03-20
Add Setting::validate_value()

wvega:
# Summary

This PR updates adds the `validate_value()` method to the `Setting` class.

## Details

The validation methods for URL and Email values use WooCommerce and WordPress functions, so instead of unit tests, I added integration tests for this method.

## QA

- [x] Unit tests pass
- [x] Integration tests pass

### DIFF
--- a/tests/integration/Settings_API/SettingTest.php
+++ b/tests/integration/Settings_API/SettingTest.php
@@ -22,4 +22,60 @@ class SettingTest extends \Codeception\TestCase\WPTestCase {
 	/** Tests *********************************************************************************************************/
 
 
+	/**
+	 * @see Setting::validate_value()
+	 *
+	 * @param mixed $value value to pass to method
+	 * @param string $type setting type
+	 * @param bool $type whether the value should be considered valid or not
+	 *
+	 * @dataProvider provider_validate_value
+	 * */
+	public function test_validate_value( $value, $type, $expected ) {
+
+		$setting = new Setting();
+		$setting->set_type( $type );
+
+		$this->assertSame( $expected, $setting->validate_value( $value ) );
+	}
+
+
+	/**
+	 * Provider for test_validate_value()
+	 *
+	 * @return array
+	 */
+	public function provider_validate_value() {
+
+		require_once( 'woocommerce/Settings_API/Setting.php' );
+
+		return [
+			[ 'example', Setting::TYPE_STRING, true ],
+			[ 3.1415926, Setting::TYPE_STRING, false ],
+
+			[ 'https://skyverge.com/', Setting::TYPE_URL, true ],
+			[ 'file:///tmp/', Setting::TYPE_URL, false ],
+			[ 'example', Setting::TYPE_URL, false ],
+
+			[ 'test@example.com', Setting::TYPE_EMAIL, true ],
+			[ 'not-an-email.com', Setting::TYPE_EMAIL, false ],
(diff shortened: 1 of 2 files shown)


tests/integration/Settings_API/SettingTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+			[ 'test@example.com', Setting::TYPE_EMAIL, true ],
+			[ 'not-an-email.com', Setting::TYPE_EMAIL, false ],
+			[ '', Setting::TYPE_EMAIL, false ],
+
+			[ 1729, Setting::TYPE_INTEGER, true ],
+			[ 'hi', Setting::TYPE_INTEGER, false ],
+
+			[ 3.14, Setting::TYPE_FLOAT, true ],
+			[ 3000, Setting::TYPE_FLOAT, false ],
```
deivamagalhaes: Should we not accept integers as options/value in a float setting?

tests/integration/Settings_API/SettingTest.php
```diff
@@ -0,0 +1,81 @@
...
+			[ 'https://skyverge.com/', Setting::TYPE_URL, true ],
+			[ 'file:///tmp/', Setting::TYPE_URL, false ],
+			[ 'example', Setting::TYPE_URL, false ],
+
+			[ 'test@example.com', Setting::TYPE_EMAIL, true ],
+			[ 'not-an-email.com', Setting::TYPE_EMAIL, false ],
+			[ '', Setting::TYPE_EMAIL, false ],
+
+			[ 1729, Setting::TYPE_INTEGER, true ],
+			[ 'hi', Setting::TYPE_INTEGER, false ],
```
deivamagalhaes: Do we want to accept numeric strings as integers? As the code is now, I believe `'1729'` would be considered valid. Just want to make sure that is what we want.

tests/integration/Settings_API/SettingTest.php
wvega: I think we can change that. I checked and if we were using type hints and `declare(strict_types=1);`, float parameters would accept both int and float values, and integer parameters would accept only ints.

Without `declare(strict_types=1);` both float and int parameters accept int, float, and numeric strings.

In either case it seems we should accept integers as options/values for float settings too. I can't make that update.

tests/integration/Settings_API/SettingTest.php
wvega: I think we don't want to accept numeric strings. So it should behave as if `declare(strict_types=1)` was used. We already force booleans to be bool, so it makes sense to do the same here.

The default implementation of `get_value_from_database()` converts values for boolean settings only, so to be strict we may need to start converting ints and floats there as well. 

I can make those updates too, if that sounds right.

tests/integration/Settings_API/SettingTest.php
deivamagalhaes: You can't? 😛

tests/integration/Settings_API/SettingTest.php
deivamagalhaes: I am good with both updates. @ChaseWiseman ?

ChaseWiseman: Sounds right to me

tests/integration/Settings_API/SettingTest.php
wvega: lol. For some reason I keep switching positives and negatives when it matters most :/

PR #446, opened 2020-03-20
Add Abstract_Settings properties & methods

wvega:
# Summary

This PR adds the properties and basic methods for the Abstract_Settings clasas.

## Details

Since `Absrtact_Settings` is... abstract, the integration tests use an an anonymous class that extends `Abstrat_Settings` as the test subject.

## QA

- [x] Unit tests pass
- [x] Integration tests pass

### DIFF
--- a/woocommerce/Settings_API/Abstract_Settings.php
+++ b/woocommerce/Settings_API/Abstract_Settings.php
@@ -260,7 +260,7 @@ abstract class Abstract_Settings {
 		/**
 		 * Filters the list of valid control types.
 		 *
-		 * @param array $setting_types valid control types
+		 * @param string[] $control_types valid control types
 		 * @param Abstract_Settings $settings the settings handler instance
 		 */
 		return apply_filters( "wc_{$this->get_id()}_settings_api_control_types", $control_types, $this );
--- a/woocommerce/Settings_API/Abstract_Settings.php
+++ b/woocommerce/Settings_API/Abstract_Settings.php
@@ -24,6 +24,8 @@
 
 namespace SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API;
 
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1 as Framework;
+
 defined( 'ABSPATH' ) or exit;
 
 if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_6_1\\Settings_API\\Abstract_Settings' ) ) :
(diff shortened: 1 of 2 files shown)


woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
  */
 abstract class Abstract_Settings {
 
+
+	/** @var string settings ID */
+	public $id = '';
```
deivamagalhaes: Do we need to initialize it here, since it will be in the constructor? 🤔

woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
...
+	 * @return string
+	 */
+	public function get_id() {
+
+		return $this->id;
+	}
+
+
+	/**
+	 * Gets all registered settings.
```
deivamagalhaes: It won't get all if you pass the IDs, right? 😬

woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
...
+	 *
+	 * @param array $ids setting IDs to get
```
deivamagalhaes: I believe we are trying to specify the array type when we can.
```suggestion
	 * @param string[] $ids setting IDs to get
```

woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
...
+		return $value;
+	}
+
+
+	/**
+	 * Gets the list of valid setting types.
+	 *
+	 * @since x.y.z
+	 *
+	 * @return array
```
deivamagalhaes:
```suggestion
	 * @return string[]
```

woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
...
+			Setting::TYPE_EMAIL,
+			Setting::TYPE_INTEGER,
+			Setting::TYPE_FLOAT,
+			Setting::TYPE_BOOLEAN,
+		];
+
+		/**
+		 * Filters the list of valid setting types.
+		 *
+		 * @param array $setting_types valid setting types
```
deivamagalhaes:
```suggestion
		 * @param string[] $setting_types valid setting types
```

woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
...
+		return apply_filters( "wc_{$this->get_id()}_settings_api_setting_types", $setting_types, $this );
+	}
+
+
+	/**
+	 * Gets the list of valid control types.
+	 *
+	 * @since x.y.z
+	 *
+	 * @return array
```
deivamagalhaes:
```suggestion
	 * @return string[]
```

woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -35,6 +37,249 @@
...
+			Control::TYPE_SELECT,
+			Control::TYPE_FILE,
+			Control::TYPE_COLOR,
+			Control::TYPE_RANGE,
+		];
+
+		/**
+		 * Filters the list of valid control types.
+		 *
+		 * @param array $setting_types valid control types
```
deivamagalhaes:
```suggestion
		 * @param string[] $control_types valid control types
```

woocommerce/Settings_API/Abstract_Settings.php
wvega: I think we don't. Fixed in 05d04f0.

woocommerce/Settings_API/Abstract_Settings.php
wvega: Good catch. I modified it in f272926.

woocommerce/Settings_API/Abstract_Settings.php
wvega: That makes sense. I committed your suggestions. Thanks!

PR #451, opened 2020-03-20
Add register_setting() to Abstract_Settings

deivamagalhaes:
# Summary

Adds the `Abstract_Settings::register_setting` method and an integration test for it.

I've kept the exception handling from the POC, because I liked it. It catches any exceptions and returns false (logging it with `wc_doing_it_wrong`). This way the plugin `register_settings` does not have to worry about catching exceptions.

## QA

- [x] Integration tests pass

### DIFF
--- a/tests/integration/Settings_API/AbstractSettingsTest.php
+++ b/tests/integration/Settings_API/AbstractSettingsTest.php
@@ -44,6 +44,35 @@ class AbstractSettingsTest extends \Codeception\TestCase\WPTestCase {
 
 
 	/**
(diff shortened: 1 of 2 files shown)


wvega:
>I've kept the exception handling from the POC, because I liked it. It catches any exceptions and returns false (logging it with wc_doing_it_wrong). This way the plugin register_settings does not have to worry about catching exceptions.

This makes sense. It mades it a little easier on users of the class. In that case we should update `delete_value()` in a similar way to return `false` instead of throwing an exception if the setting doesn't exist.

PR #458, opened 2020-03-23
Add update_value() to Abstract_Settings

deivamagalhaes:
# Summary

Adds the `get_value` method to the handler class.

## QA

- [x] Unit tests pass
- [x] Integration tests pass

### DIFF
--- a/woocommerce/Settings_API/Abstract_Settings.php
+++ b/woocommerce/Settings_API/Abstract_Settings.php
@@ -312,6 +312,8 @@ abstract class Abstract_Settings {
 
 		// performs the validations and updates the value
 		$setting->update_value( $value );
+
+		$this->save( $setting_id );
 	}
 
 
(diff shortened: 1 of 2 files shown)


woocommerce/Settings_API/Abstract_Settings.php
```diff
@@ -200,6 +200,28 @@ public function get_setting( $id ) {
...
+	public function update_value( $setting_id, $value ) {
+
+		$setting = $this->get_setting( $setting_id );
+
+		if ( ! $setting ) {
+			throw new Framework\SV_WC_Plugin_Exception( "Setting {$setting_id} does not exist" );
+		}
+
+		// performs the validations and updates the value
+		$setting->update_value( $value );
```
wvega: Should we also call `$this->save( $setting_id )` here to update the option as well?

deivamagalhaes: I am not sure. I'll add a question to the story!

deivamagalhaes: Actually, since `delete_value` deletes the value from the DB, that makes sense. Thanks for pointing it out.

PR #459, opened 2020-03-23
Add the REST API's prepare_item_for_response() method

wvega:
# Summary

This PR adds the `prepare_item_for_response()` method to the `Settings` REST Controller class.

## Details

I wasn't sure whether the implementation document expected the method to return an array, or a REST Response object containing an array. However, since `WP_REST_Controller::prepare_item_for_response()` returns `WP_Error|WP_REST_Response`, I decided to match the parent's method return type in this PR.

From what I can see, we would be converting the array into a response object anyway, so I think it makes sense to match the return type of the parent method.

## QA

- [x] Integration tests pass

### DIFF
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -84,6 +84,20 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 	}
 
 
+	/** @see Abstract_Settings::get_id() */
+	public function test_prepare_item_for_response_value_not_set() {
+
+		$settings   = $this->get_settings_instance();
+		$controller = new Settings( $settings );
+
+		$settings->register_setting( 'test', Setting::TYPE_STRING );
+
+		$item = $controller->prepare_item_for_response( $settings->get_setting( 'test' ), null );
+
+		$this->assertEquals( null, $item['value'] );
+	}
+
+
 	/** Helper methods ************************************************************************************************/
 
 
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -94,6 +94,8 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 
 		$item = $controller->prepare_item_for_response( $settings->get_setting( 'test' ), null )->get_data();
 
+		// $item['value'] is null if no value has been set for the setting
+		// we want users of the API to differentiate between a setting that's been set vs. a setting that has a default value but not saved yet
 		$this->assertEquals( null, $item['value'] );
 	}
 
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -1,6 +1,9 @@
 <?php
 
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\REST_API\Controllers\Settings;
 use SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API\Abstract_Settings;
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API\Setting;
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API\Control;
 
 /**
  * Tests for the Abstract_Settings class.
@@ -20,7 +23,10 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 
 	protected function _before() {
 
+		require_once 'woocommerce/rest-api/Controllers/Settings.php';
 		require_once 'woocommerce/Settings_API/Abstract_Settings.php';
+		require_once 'woocommerce/Settings_API/Control.php';
+		require_once 'woocommerce/Settings_API/Setting.php';
 	}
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
...
+	public function test_prepare_item_for_response_value_not_set() {
+
+		$settings   = $this->get_settings_instance();
+		$controller = new Settings( $settings );
+
+		$settings->register_setting( 'test', Setting::TYPE_STRING );
+
+		$item = $controller->prepare_item_for_response( $settings->get_setting( 'test' ), null )->get_data();
+
+		$this->assertEquals( null, $item['value'] );
```
wvega: I'm thinking that we should return the default value instead, as in `Abstract_Settings::get_value()`. What do you think?

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
...
+		$settings->register_setting( 'test', Setting::TYPE_STRING, [
+			'name'        => 'Test Setting',
+			'description' => 'A simple setting',
+			'options'     => [ 'a', 'b', 'c' ],
+			'default'     => 'c',
+		] );
+
+		$settings->register_control( 'test', Control::TYPE_SELECT, [
+			'name'  => 'Select field',
+			'description' => 'A regultar select input field',
```
deivamagalhaes:
```suggestion
			'description' => 'A regular select input field',
```

tests/integration/REST_API/Controllers/SettingsTest.php
deivamagalhaes: I am not sure. If we do, we would lose visibility from the API if the value is set or not. Is there any circumstance when we would need to know if the database option is not set or is set to the default value? @ChaseWiseman , thoughts?

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -155,7 +156,44 @@ public function update_item_permissions_check( $request ) {
 	/** Utility methods ***********************************************************************************************/
 
 
-	// TODO: prepare_item_for_response()
+	/**
+	 * @sincex.y.z
```
deivamagalhaes:
```suggestion
	 * @since x.y.z
```

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -155,7 +156,44 @@ public function update_item_permissions_check( $request ) {
...
+				'name'        => $setting->get_name(),
+				'description' => $setting->get_description(),
+				'is_multi'    => $setting->is_is_multi(),
+				'options'     => $setting->get_options(),
+				'default'     => $setting->get_default(),
+				'value'       => $setting->get_value(),
+			];
+
+			if ( $control = $setting->get_control() ) {
+				$item['control'] = [
```
deivamagalhaes: I think it may be better for consistency to have the control key with a null value then not to have it.

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -155,7 +156,44 @@ public function update_item_permissions_check( $request ) {
 	/** Utility methods ***********************************************************************************************/
 
 
-	// TODO: prepare_item_for_response()
+	/**
+	 * @sincex.y.z
+	 *
+	 * @param Setting $setting a setting object
+	 * @param \WP_REST_Request $request request object
+	 * @return WP_Error|WP_HTTP_Response
```
deivamagalhaes: Shouldn't this be `WP_REST_Response`?

deivamagalhaes:
```suggestion
	 * @return WP_Error|WP_REST_Response
```

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+	protected function _after() {
+
+	}
+
+
+	/** Tests *********************************************************************************************************/
+
+
+	/** @see Abstract_Settings::get_id() */
```
deivamagalhaes: Is this correct?

tests/integration/REST_API/Controllers/SettingsTest.php
ChaseWiseman: I think in the REST context, it'd be valuable to differentiate between a setting that's been explicitly set vs. a setting that may have a default but not yet saved. The consumer (e.g. JS app) can decide whether they want to pre-fill their field with the default value or display/use it in some other way.

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
...
+		$this->assertEquals( $setting->get_value(),       $item['value'] );
+
+		$this->assertEquals( $control->get_type(),        $item['control']['type'] );
+		$this->assertEquals( $control->get_name(),        $item['control']['name'] );
+		$this->assertEquals( $control->get_description(), $item['control']['description'] );
+		$this->assertEquals( $control->get_options(),     $item['control']['options'] );
+	}
+
+
+	/** @see Abstract_Settings::get_id() */
```
deivamagalhaes: Same here.

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+		$settings->register_setting( 'test', Setting::TYPE_STRING );
+
+		$item = $controller->prepare_item_for_response( $settings->get_setting( 'test' ), null )->get_data();
+
+		$this->assertEquals( null, $item['value'] );
+	}
+
+
+	/** @see Abstract_Settings::get_id() */
```
deivamagalhaes: Same here

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
+<?php
+
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\REST_API\Controllers\Settings;
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API\Abstract_Settings;
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API\Setting;
+use SkyVerge\WooCommerce\PluginFramework\v5_6_1\Settings_API\Control;
+
+/**
+ * Tests for the Abstract_Settings class.
```
deivamagalhaes: This one also needs to be updated.

tests/integration/REST_API/Controllers/SettingsTest.php
wvega: Noup. I left lots of copypasta on this class :(. Fixed in a717049

woocommerce/rest-api/Controllers/Settings.php
wvega: Yes. Good catch!

woocommerce/rest-api/Controllers/Settings.php
wvega: Sounds good. I changed that in 6249c31 and 5499ec1

wvega:
@deivamagalhaes thank you for reviewing. I'm sorry for all the copypasta and typos 🙏

I think I already addressed all your feedback. Would you mind taking another look?

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -0,0 +1,142 @@
...
+	public function test_prepare_item_for_response_value_not_set() {
+
+		$settings   = $this->get_settings_instance();
+		$controller = new Settings( $settings );
+
+		$settings->register_setting( 'test', Setting::TYPE_STRING );
+
+		$item = $controller->prepare_item_for_response( $settings->get_setting( 'test' ), null )->get_data();
+
+		$this->assertEquals( null, $item['value'] );
```
deivamagalhaes: Maybe a comment above this line to document our decision?

wvega: Good call. I added one in fa6bfdb.

PR #460, opened 2020-03-24
Update base REST handler to add a settings controller & endpoints

wvega:
# Summary

This PR updates the `register_routes()` method to the `REST_API` REST handler.

## Details

The test for `register_routes()` checks for the existence of one of the routes that the `REST_API\Controllers\Settings` class is supposed to register as proof that the `REST_API\Controllers\Settings::register_routes()` method was called. We should write separate tests for the controller class to verify that all the necessary routes are registered.

## QA

- [ ] Unit tests pass
- [ ] Integration tests pass

### DIFF
--- a/woocommerce/rest-api/Controllers/Settings.php
+++ b/woocommerce/rest-api/Controllers/Settings.php
@@ -51,10 +51,9 @@ class Settings extends \WP_REST_Controller {
 	 */
 	public function __construct( Abstract_Settings $settings ) {
 
-		$this->settings = $settings;
+		$this->settings  = $settings;
 		$this->namespace = 'wc/v3';
-
-		// TODO: set $this->rest_base when Abstract_Settings has a get_id() method
+		$this->rest_base = "{$settings->get_id()}/settings";
 	}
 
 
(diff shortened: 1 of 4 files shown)


woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -51,10 +51,9 @@ class Settings extends \WP_REST_Controller {
 	 */
 	public function __construct( Abstract_Settings $settings ) {
 
-		$this->settings = $settings;
+		$this->settings  = $settings;
 		$this->namespace = 'wc/v3';
-
-		// TODO: set $this->rest_base when Abstract_Settings has a get_id() method
+		$this->rest_base = "{$settings->get_id()}/settings";
```
wvega: Is this the endpoint we want? I added a question about that to the [story](https://app.clubhouse.io/skyverge/story/32703/add-rest-handler-class#activity-33669) were we defined the initial version of the class.

deivamagalhaes: LGTM, but let's wait and see if @ChaseWiseman has any comments on it...

PR #462, opened 2020-03-24
Add the REST API get_item() method

deivamagalhaes:
# Summary

Adds `get_item` method.

## Details

I've added a few comments/questions for things I am not sure.

## QA

- [x] Unit tests pass
- [x] Integration tests pass

### DIFF
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -95,6 +95,25 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 	}
 
 
+	/** @see Settings::get_item() */
+	public function test_get_item_not_found() {
+
+		$settings   = $this->get_settings_instance();
+		$controller = new Settings( $settings );
+
+		$setting_id = 'test';
+
+		$request = new WP_REST_Request( 'GET', "/wc/v3/{$settings->get_id()}/settings" );
+		$request->set_url_params( [ 'id' => $setting_id ] );
+
+		$response = $controller->get_item( $request );
+
+		$this->assertTrue( $response instanceof WP_Error );
+		$this->assertSame( 'wc_rest_setting_not_found', $response->get_error_code() );
+		$this->assertSame( [ 'status' => 404 ], $response->get_error_data() );
+	}
+
+
 	/** @see Settings::prepare_item_for_response() */
 	public function test_prepare_item_for_response() {
 
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -55,7 +55,7 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 
 		$settings->register_control( $setting_id, Control::TYPE_SELECT, [
 			'name'        => 'Select field',
-			'description' => 'A regultar select input field',
+			'description' => 'A regular select input field',
 			'options'     => [
 				'a' => 'A',
 				'b' => 'B',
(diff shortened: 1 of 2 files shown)


tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -38,6 +38,82 @@ protected function _after() {
...
+			],
+		] );
+
+		$setting = $settings->get_setting( $setting_id );
+		$control = $setting->get_control();
+
+		$setting->set_value( 'a' );
+
+		$request = new WP_REST_Request( 'GET', "/wc/v3/{$settings->get_id()}/settings" );
+		$request->set_url_params( [ 'id' => $setting_id ] );
```
deivamagalhaes: I am not 100% sure this is the correct way to send the param.

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -125,7 +125,35 @@ public function get_items_permissions_check( $request ) {
...
+	 * @param \WP_REST_Request $request request object
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public function get_item( $request ) {
+
+		$setting_id = $request->get_param( 'id' );
+
+		if ( $setting = $this->settings->get_setting( $setting_id ) ) {
+
+			return $this->prepare_item_for_response( $setting, $request );
```
deivamagalhaes: No need to call `rest_ensure_response` because `prepare_item_for_response` already calls it.

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -125,7 +125,35 @@ public function get_items_permissions_check( $request ) {
...
+		} else {
+
+			return new \WP_Error(
+				'wc_rest_setting_not_found',
+				sprintf(
+					/* translators: Placeholder: %s - setting ID */
+					__( 'Setting %s does not exist', 'woocommerce-plugin-framework' ),
+					$setting_id
+				),
+				[ 'status' => 404 ]
```
deivamagalhaes: Is this how we want to set the status? Will it translate to REST status? I am still trying to make the API work in Postman so I can check.

wvega: Yes, I think that should work. See [`class-wp-rest-server.php#L159-L165`](https://github.com/WordPress/WordPress/blob/55482506a3cba25dbef5930d25b09611029c3264/wp-includes/rest-api/class-wp-rest-server.php#L159-L165).

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -38,6 +38,63 @@ protected function _after() {
...
+				'c' => 'C'
+			],
+		] );
+
+		$setting = $settings->get_setting( $setting_id );
+		$control = $setting->get_control();
+
+		$setting->set_value( 'a' );
+
+		$request = new WP_REST_Request( 'GET', "/wc/v3/{$settings->get_id()}/settings" );
```
wvega:
```suggestion
		$request = new WP_REST_Request( 'GET', "/wc/v3/{$settings->get_id()}/settings/{$setting_id}" );
```

The route for a single item should include the ID of the setting in the URL. However, this doesn't change anything in the test, because the request object is not the one that extracts the parameters from the URL.

tests/integration/REST_API/Controllers/SettingsTest.php
wvega: I think that's how it's done internally.

tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -95,6 +95,25 @@ public function test_get_item() {
...
 
+	/** @see Settings::get_item() */
+	public function test_get_item_not_found() {
+
+		$settings   = $this->get_settings_instance();
+		$controller = new Settings( $settings );
+
+		$setting_id = 'test';
+
+		$request = new WP_REST_Request( 'GET', "/wc/v3/{$settings->get_id()}/settings" );
```
wvega:
```suggestion
		$request = new WP_REST_Request( 'GET', "/wc/v3/{$settings->get_id()}/settings/{$setting_id}" );
```

PR #464, opened 2020-03-25
Add the REST API get_items() method

wvega:
# Summary

This PR adds the `get_items()` method to the Settings REST controller.

## Details

I also added a couple of helper methods to assert that an item's data matches the properties of a given setting or control object. We are performing those verifications in several tests so it made sense to me to reuse the code.

## QA

- [x] Unit tests pass
- [x] Integration tests pass

### DIFF
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -221,24 +221,7 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 		$settings   = $this->get_settings_instance();
 		$controller = new Settings( $settings );
 
-		$settings->register_setting( 'test', Setting::TYPE_STRING, [
-			'name'        => 'Test Setting',
-			'description' => 'A simple setting',
-			'options'     => [ 'a', 'b', 'c' ],
-			'default'     => 'c',
-		] );
-
-		$settings->register_control( 'test', Control::TYPE_SELECT, [
-			'name'  => 'Select field',
-			'description' => 'A regular select input field',
-			'options'     => [
-				'a' => 'A',
-				'b' => 'B',
-				'c' => 'C'
-			],
-		] );
-
-		$setting = $settings->get_setting( 'test' );
+		$setting = $settings->get_setting( 'test_one' );
 		$setting->set_value( 'a' );
 
 		$item = $controller->prepare_item_for_response( $setting, null )->get_data();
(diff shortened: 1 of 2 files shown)


tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -38,46 +81,50 @@ protected function _after() {
 	/** Tests *********************************************************************************************************/
 
 
-	/** @see Settings::get_item() */
-	public function test_get_item() {
```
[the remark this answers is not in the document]
deivamagalhaes: Oh, it got moved to after the helper methods... I believe it would be better if it was right after `test_get_items`. And the helpers after all the test methods.

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -122,7 +122,30 @@ public function get_items_permissions_check( $request ) {
...
+	 */
+	public function get_items( $request ) {
+
+		$items = [];
+
+		foreach ( $this->settings->get_settings() as $setting ) {
+
+			$response = $this->prepare_item_for_response( $setting, $request );
+
+			// check that the response object has data
```
deivamagalhaes: Hmmm, maybe we should move `rest_ensure_response` out of the `prepare_item_for_response` at this point? It feels weird to wrap the data array in a response object and then extract the array from it... What do you think?

tests/integration/REST_API/Controllers/SettingsTest.php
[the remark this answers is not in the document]
wvega: Sounds good. I moved the helpers to the bottom in e0fabbe.

woocommerce/rest-api/Controllers/Settings.php
wvega: I agree, it feels a bit weird. Since `prepare_item_for_response()` is defined in the parent class, I think we should also rename `prepare_item_for_response()` to, for example, `prepare_setting_item()`, and return the array version there.

What do you think?

deivamagalhaes: Sounds good to me. Maybe add a note to the original story documenting the decision.

wvega: Sounds good. I modified the method in 615ef5d and added a note to the [story](https://github.com/skyverge/wc-plugin-framework/pull/464#discussion_r397551765).

PR #465, opened 2020-03-25
Add the REST API update_item() method

wvega:
# Summary

This PR adds the `update_item()` method to the Settings REST controller class.

## Details

The method enhances the `settings/{setting_id}` endpoint to allow users to update the value of the setting. Since we perform type validation on `Setting::update_value()`, updating the value of a setting only works if the body of the request is a JSON object with the `value` property set to the new value using the appropriate type for the setting.

## QA

- [x] Unit tests pass
- [x] Integration tests pass

### DIFF
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -184,6 +184,42 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 	}
 
 
+	/**
+	 * @see Settings::update_item()
+	 *
+	 * @param string $setting_id setting ID
+	 * @param array|bool|float|int|string $value setting value
+	 *
+	 * @dataProvider provider_test_update_item_error
+	 */
+	public function test_update_item_error( $setting_id, $value ) {
+
+		$settings   = $this->get_settings_instance();
+		$controller = new Settings( $settings );
+		$request    = new WP_REST_Request( 'POST', "/wc/v3/{$settings->get_id()}/settings/{$setting_id}" );
+
+		$request->set_header( 'content-type', 'application/json' );
+		$request->set_url_params( [ 'id' => $setting_id ] );
+		$request->set_body( json_encode( [ 'value' => $value ] ) );
+
+		$response = $controller->update_item( $request );
+
+		$this->assertInstanceOf( WP_Error::class, $response );
+		$this->assertSame( 'wc_rest_setting_could_not_update', $response->get_error_code() );
+		$this->assertSame( [ 'status' => 400 ], $response->get_error_data() );
+	}
+
+
+	/** @see test_update_item_error() */
+	public function provider_test_update_item_error() {
+
+		return [
+			'setting not found' => [ 'not_found', null ],
+			'invalid value'     => [ 'test_one', 1234 ],
+		];
+	}
+
+
 	/** @see Settings::prepare_setting_item() */
 	public function test_prepare_setting_item() {
 
--- a/tests/integration/REST_API/Controllers/SettingsTest.php
+++ b/tests/integration/REST_API/Controllers/SettingsTest.php
@@ -162,6 +162,28 @@ class SettingsTest extends \Codeception\TestCase\WPTestCase {
 	}
 
 
+	/** @see Settings::update_item() */
+	public function test_update_item() {
+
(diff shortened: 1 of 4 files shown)


tests/integration/REST_API/Controllers/SettingsTest.php
```diff
@@ -162,6 +162,64 @@ public function test_get_item_not_found() {
...
+		$request->set_header( 'content-type', 'application/json' );
+		$request->set_url_params( [ 'id' => $setting->get_id() ] );
+		$request->set_body( json_encode( [ 'value' => 'a' ] ) );
+
+		$response = $controller->update_item( $request );
+
+		$this->assertInstanceOf( WP_REST_Response::class, $response );
+		$this->assertSame( 200, $response->get_status() );
+
+		$this->assert_item_matches_setting( $response->get_data(), $setting );
```
deivamagalhaes: Should we maybe also test if the value was really updated in the database, by getting it again?

woocommerce/rest-api/Controllers/Settings.php
```diff
@@ -194,13 +194,47 @@ public function update_item_permissions_check( $request ) {
...
+			$value      = $request->get_param( 'value' );
+
+			// throws an exception if the setting doesn't exist or the value is not valid
+			$this->settings->update_value( $setting_id, $value );
+
+			return rest_ensure_response( $this->prepare_setting_item( $this->settings->get_setting( $setting_id ), $request ) );
+
+		} catch ( \Exception $e ) {
+
+			return new \WP_Error(
```
deivamagalhaes: I'm not sure if that is what @ChaseWiseman meant by `Exception codes are passed to any \WP_Error` when he wrote the acceptance criteria, but should we maybe return different errors for each case (not found or invalid value)? We would have to update the `Abstract_Settings::update_value` and `Setting::update_value` to throw different exceptions. But it would allow the UI using this API to display specific error messages... 🤔

wvega:
I added another commit to use the exception code as the response status if code was provided (
e1f3a6e).

woocommerce/rest-api/Controllers/Settings.php
wvega: @deivamagalhaes you make a good point. The errors would be more useful that way.

You mean changing `wc_rest_setting_could_not_update` to `wc_rest_setting_not_found` and `wc_rest_setting_invalid_value` depending on the type of exception?

If we add `400` and `404` codes to the exception thrown when the value is invalid and when the setting doesn't exist, we may be able to check the code on the `catch` block and adjust the error code and message accordingly.
Do we normally use the exceptions codes that way?

I think I prefer to have two new exceptions for these specific cases, but we also don't have many subclasses of `SV_WC_Payment_Gateway_Exception` either?

What do you think?

tests/integration/REST_API/Controllers/SettingsTest.php
wvega: Makes sense. I retrieved and compared the value stored in the associated option in 623373c. Also added an assert to confirm the returned value matches the one we sent.

woocommerce/rest-api/Controllers/Settings.php
deivamagalhaes: > You mean changing wc_rest_setting_could_not_update to wc_rest_setting_not_found and wc_rest_setting_invalid_value depending on the type of exception?

Yes, this would be what I would prefer. 

But maybe this point is something we should discuss with the whole team?

ChaseWiseman: @deivamagalhaes @wvega sorry for being a bit vague there!

I meant the code we set for the `\WP_Error`'s `'status'` value. That can be set using `$e->getCode()`, and whenever throwing them in the related code we'd pass the appropriate HTTP code depending on the problem. `404` for not found,`400` for invalid, etc... whatever we set there ends up as the HTTP response code.

I think the _string_ error code, `wc_rest_setting_could_not_update` can remain the same either way

wvega: Thank you @deivamagalhaes @ChaseWiseman.

I updated `Abstract_Settigns::update_value()` and `Setting::update_value()` to add the appropriate error code (`404` or `400`) to the exceptions they throw (9f8cae5 and 269def0) and modified the tests to check those values (1abd1c1).

Also updated the story description with a link to this conversation.

PR #470, opened 2020-03-27
Epic: Improvements for JS handling in registered scripts

wvega:
This is the release branch for the [Improvements for JS handling in registered scripts epic](https://app.clubhouse.io/skyverge/epic/30527/improvements-for-js-handling-in-registered-scripts).


woocommerce/payment-gateway/Frontend/Script_Handler.php
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * Returns the JS handler class name.
+	 *
+	 * @since x.y.z
+	 *
+	 * @return string
+	 */
+	protected function get_js_handler_class_name() {
+
+		return sprintf( '%s_5_6_1', $this->js_handler_base_class_name );
```
unfulvio:
```suggestion
		return sprintf( '%s_v5_6_1', $this->js_handler_base_class_name );
```

hey guys @ChaseWiseman @wvega @deivamagalhaes just had an afterthought about the inline JS namespace containing the script version adapted from the framework semver. Should we prepend a `v` here? So it matches the namespace in PHP classes e.g. `SkyVerge\WooCommerce\PluginFramework\v5_6_1\Frontend`; concern is just if we mass rename these what if we are doing search & replace with `v5_6_1` rather than `5_6_1` as the search needle?

deivamagalhaes: Adjusted in https://github.com/skyverge/wc-plugin-framework/pull/475

PR #475, opened 2020-03-28
Trigger a script loaded event in payment gateway JavaScript handlers

deivamagalhaes:
# Summary

This PR dispatches the loaded events for the JS handlers after defining each class.

### Setup

- Update `composer.json` in Authorize.Net to use this branch (`dev-ch33803/loaded-event`)
- Configure Authorize.Net in test mode (both Credit Card and ECheck)
- Update `woocommerce-gateway-authorize-net-cim/includes/Payment_Form.php` replacing the entire `render_js()` method with these two methods:
    ```php
	/**
	 * Gets the JS args for the payment form handler.
	 *
	 * @since x.y.z
	 *
	 * @return array
	 */
	protected function get_js_handler_params() {

		$args = array_merge( parent::get_js_handler_params(), [
			'logging_enabled'  => $this->get_gateway()->debug_log(),
			'lightbox_enabled' => $this->get_gateway()->is_lightbox_payment_form_enabled(),
			'login_id'         => $this->get_gateway()->get_api_login_id(),
			'client_key'       => $this->get_gateway()->get_client_key(),
			'general_error'    => __( 'An error occurred, please try again or try an alternate form of payment.', 'woocommerce-gateway-authorize-net-cim' ),
			'ajax_url'         => admin_url( 'admin-ajax.php' ),
			'ajax_log_nonce'   => wp_create_nonce( 'wc_' . $this->get_gateway()->get_id() . '_log_js_data' ),
		] );

		return $args;
	}


	/**
	 * Returns the JS handler class name.
	 *
	 * @since x.y.z
	 *
	 * @return string
	 */
	protected function get_js_handler_class_name() {

		return 'WC_Authorize_Net_Payment_Form_Handler';
	}
    ```
- Activate Apple Pay for Authorize.Net using this code snippet:
    ```php
    add_filter( 'wc_payment_gateway_authorize_net_cim_activate_apple_pay', '__return_true' );
    ````
- Make sure Apple Pay is allowed on Single Product pages
- Set **Apple Merchant ID** to `merchant.com.skyverge.test`
    That's a fake ID, but I think we don't need Apple Pay to really work to confirm the handler is correctly initialized.
- Point **Certificate Path** to any file on your website's root directory

#### Payment Form

1. Update Authorize.Net `SV_WC_Payment_Gateway_Payment_Form` adding the following line to the JS enqueued in `render_js`:
```
window.jQuery( document.body ).on( '<?php echo esc_js( $loaded_event ); ?>', console.log('loaded payment form') );
```
1. Add a product to the cart and go to the Checkout page
1. Open the browser's console
    - [x] `loaded payment form` is displayed twice (one for Credit Cart and one for ECheck)

#### Payment Methods

1. Update Authorize.Net `SV_WC_Payment_Gateway_My_Payment_Methods` adding the following line to the JS enqueued in `render_js`:
```
window.jQuery( document.body ).on( '<?php echo esc_js( $loaded_event ); ?>', console.log('loaded payment methods') );
```
1. Go to My Account > Payment Methods
1. Open the browser's console
    - [x] `loaded payment methods` is displayed once

#### Apple Pay

1. Update Authorize.Net `SV_WC_Payment_Gateway_Apple_Pay_Frontend` adding the following line to the JS enqueued in `enqueue_js_handler`:
```
window.jQuery( document.body ).on( '<?php echo esc_js( $loaded_event ); ?>', console.log('loaded apple pay') );
```
1. Go to a product's page
1. Open the browser's console
    - [x] `loaded apple pay` is displayed once


woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
```diff
@@ -382,3 +382,7 @@ jQuery( document ).ready ($) ->
 		#
 		# @since 3.0.0
 		unblock_ui: -> @form.unblock()
+
+
+	# dispatch loaded event
+	$( document.body ).trigger( "wc_#{ @plugin_id }_payment_form_handler_loaded" )
```
deivamagalhaes: @wvega, I got the event name from the page source, but I thought it was a bit weird that this event was different from the others. Is it correct?

wvega: The `wc_authorize_net_payment_form_handler_loaded` event name in this case is a side effect of the overwritten `get_js_handler_class_name()` method in `woocommerce-gateway-authorize-net-cim/includes/Payment_Form.php`. 

I believe it's correct that `render_js()` is using that event name because Authorize.Net initializes a subclass of the handler included in the framework (`WC_Authorize_Net_Payment_Form_Handler`) that needs to be loaded separately.

However, `sv-wc-payment-gateway-payment-form.coffee` should probably still trigger `sv_wc_payment_form_X_Y_Z_loaded` so that it works on gateways that use the base handler directly.

In order to upgrade Authorize.Net to use this version of the framework, we would also need to trigger `wc_authorize_net_payment_form_handler_loaded` in `woocommerce-gateway-authorize-net-cim/assets/js/frontend/wc-authorize-net-cim.coffee`.

deivamagalhaes: That makes sense, I've fixed the event for this branch.

wvega:
@unfulvio @deivamagalhaes both suggestions make sense to me. It may be a little easier to replace namespaces during framework releases if we use the same pattern `vX_Y_Z`.

>The second suggestion si to add a public helper method for getting the `$loaded_event`.

Perhaps the method can be `protected` as `get_js_handler_params()` and `get_js_handler_class_name()`. 

If we agree, I suggest to update the event names added in this PR and create another story for adding the new method and updating `get_js_handler_class_name()`.

deivamagalhaes:
I've created https://app.clubhouse.io/skyverge/story/48434/update-namespaced-js-classes-to-include-the-v-prefix to update `get_js_handler_class_name`. I'll update the event and add the new method on this PR.

woocommerce/payment-gateway/Frontend/Script_Handler.php
```diff
@@ -54,7 +54,20 @@ abstract class Script_Handler {
 	 */
 	protected function get_js_handler_class_name() {
 
-		return sprintf( '%s_5_6_1', $this->js_handler_base_class_name );
+		return sprintf( '%s_v5_6_1', $this->js_handler_base_class_name );
+	}
+
+
+	/**
+	 * Returns the JS event triggered after the JS handler class is loaded.
```
unfulvio, line 62:
```suggestion
	 * Gets the JS event triggered after the JS handler class is loaded.
```

should prefer `Gets` over `Returns`? I recall some discussion in the past about the verb to use - we voted for gets, but I can't find a reference in the wiki, sorry

PR #477, opened 2020-03-30
Adds an AJAX request to log script errors

unfulvio:
## Summary

Adds a method to the script handler abstract to output some inline JavaScript to issue an AJAX post request to track script errors.

## Additional details

See #478 for callback implementation.

The added method presumes a `get_plugin()` method to be available in order to be able to extract the plugin ID. This can be added in #478.

## QA

- [x] You can test this by implementing this version of the framework in a child plugin. The plugin should output the corresponding JS code inline at payment form, apple form, payment methods screen.

### DIFF
--- a/woocommerce/payment-gateway/Frontend/Script_Handler.php
+++ b/woocommerce/payment-gateway/Frontend/Script_Handler.php
@@ -73,14 +73,27 @@ abstract class Script_Handler {
 		ob_start();
 
 		?>
+
+		var errorName    = '',
+		    errorMessage = '';
+
+		if ( undefined === err || 0 === err.length ) {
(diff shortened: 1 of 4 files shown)


woocommerce/payment-gateway/Frontend/Script_Handler.php
```diff
@@ -56,6 +58,48 @@ protected function get_js_handler_class_name() {
 	}
 
 
+	/**
+	 * Gets inline JavaScript code to issue an AJAX request to add a script error event to the debug log.
```
deivamagalhaes:
```suggestion
	 * Gets inline JavaScript code to issue an AJAX request to log a script error event.
```

PR #478, opened 2020-03-30
Adds a callback to AJAX requests to log script errors 

unfulvio:
## Summary

Adds an action callback that listens to AJAX requests to log a script error.

## Additional details

Unfortunately I've hit a small issue with this that prompted me to change the tasks implementation as detailed in the story. 

Seems that for at least the Payment Form handler, the callbacks we originally wanted to include in the `Frontend\Script_Handler` would not trigger early enough to catch the AJAX request and log the error. The Payment Form handler is output in frontend during template output time, so a bit too late and I could not see a way to include the AJAX handling directly inside `Frontend\Script_Handler`  and initializing the Payment Form early enough without running into risks of having certain hooks firing twice inside that handler, or other unnecessary refactorings. Instead, I placed the AJAX request callback inside the gateway main file, along with other log handling functions. This seems to work. 

Other details in this [Slack thread](https://skyverge.slack.com/archives/CV25U34GZ/p1585597377017300).

This issue was a bit difficult to catch early on as timing issues may not be easy to be seen, especially within abstractions presented by the framework.

## QA

You can test this by attaching the PR to a gateway plugin, then reach the screens where the script would appear. If there's an error (you can purposely trigger one), then an AJAX request would be triggered. If the callback has registered early enough, the issue would be logged in the plugin's main log.

Example log result you should see:

image


unfulvio:
> It works as expected, but it took me a while to find the logs. Wouldn't it be better to log these in the gateway log, instead of the plugin's?

Hey @deivamagalhaes the gateways logs seem to record mostly API issues, I was thinking this sort of issue could be better highlighted in the main plugin log (which is mostly underutilized in gateways), so it may be easier for merchants or support team to spot. 

However you brought up a valid point. 

In the `Frontend\Script_Handler` there might be no gateway/id context. That's why I opted for the plugin ID, since some of those classes may have a `get_plugin()` method, but not necessarily a gateway for context. So the AJAX action is `wc_<plugin_id>_log_script_event`. 

I'm wondering now whether this should be `wc_<plugin_id>_log_gateway_script_event` instead, as plugins could theoretically have other front end script facing similar issues and they might want to adopt a similar pattern.

As such, we could log the gateway related script errors in the gateway debug logs.

Or, keep it as it is, because it may be easier to spot if it's not mixed up with other gateway logging. Probably if someone sees issues coming from the gateway they might turn gateway debug logging, and the script error information get lost in the  more verbose gateway API logs. 

Thoughts? CC @wvega @ChaseWiseman

wvega:
@unfulvio @deivamagalhaes @ChaseWiseman Since `Frontend\Script_Handler` does not always has a gateway context, using the plugin ID for the event seems better to me.

For the same reason I would keep the event name as `wc_<plugin_id>_log_script_event`. The errors logged from `SV_WC_Payment_Methods_Handler`, even though gateway related in the end, use a plugin context too.

Something I just noticed: I think that if we register the Ajax handler in `SV_WC_Payment_Gateway::__construct()`, the handler is registered more than once on plugins that have more than one gateway. Wouldn't that cause the event to be logged multiple times when a script issues the Ajax call?

If we continue to use the plugin ID, perhaps it makes sense to move the handler to the plugin level too. In that case, I'm ok with storing the error on the plugin log instead of the gateway log.

ChaseWiseman:
@unfulvio @deivamagalhaes @wvega I think we may have a few too many workarounds here for dealing with the different form types and plugins vs. gateways, and we could simplify and abstract things a bit.

The fundamental things the script handler needs in order to log via AJAX are:
- A unique string for the `action` value
- A unique string for the nonce/security value
- A known method for logging in the server-side callback, e.g. `->log()`

The above things don't need to be concerned about where they're coming from, whether its a plugin or gateway, etc...

So what if we do the following:
1. Add new abstract methods:
    - `Script_Handler::get_id()`
    - `Script_Handler::get_id_dasherized()`
1. Use these methods in `get_js_handler_event_debug_log_request()` to build the AJAX request data
1. The concrete script handlers can define what the new ID methods return, whether it be a plugin, gateway, or some other ID. They represent the server-side ID for that particular script handler, no matter what it is.
This allows all of the logging JS to live in the abstract and not be aware of what's controlling it.

To @wvega's point, we'll need to avoid double logging that happens if the hooks are in the gateway class.

I think we can still add these hooks too `Script_Handler`. As already noted, `SV_WC_Payment_Gateway_Payment_Form` is [loaded a bit too late](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php#L717-L740) but we can tweak that. Like all of the framework's other handlers, we should be safe to _instantiate_ it earlier, then call `->render()` as we do now.

So if that holds true, then we can:
1. Add the AJAX logging callbacks to `Script_Handler` using the `get_id()` method to match the action names
1. Add an abstract `Script_Handler::log()` method and fire it in the log callback
1. Child classes can define `::log()` to log however they see fit, plugin or gateway.

I think with all of that above, the `Script_Handler` stays clean and has a confined set of tasks, and each concrete handler that extends it can define those few methods to route the logging where it needs to go.

I feel like there's some value in keeping the logged errors in their respective places. My Payment Methods JS does make sense in the overall plugin log, since it encompasses all gateways, but the payment form and Apple Pay scripts are gateway-specific, and I feel like errors from those should show up accordingly.

What do y'all think? If that sounds okay, I'll get a story going.

wvega:
@ChaseWiseman @unfulvio @deivamagalhaes that sounds good to me!

deivamagalhaes:
@ChaseWiseman, sounds good.

unfulvio:
closing this in favor of #488

PR #495, opened 2020-05-16
Fix gateway payment form i18n strings

ChaseWiseman:
# Summary

Fixes an issue introduced in #494 which caused the payment form parameters to not load when a single gateway is active.

## UI Changes

N/A

## QA

- You can use this First Data branch to test: https://github.com/skyverge/wc-plugins/pull/3717
- Enable the `Payeezy Gateway` mode

1. Load checkout
    - [x] The payment form loads without error
1. Clear the CSC field and try to submit the order
    - [x] The JS handles the field and a missing CSC message is shown

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -443,16 +443,17 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 			return;
 		}
 
-		$handle = 'sv-wc-payment-gateway-payment-form';
+		$handle           = 'sv-wc-payment-gateway-payment-form';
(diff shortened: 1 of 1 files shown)


ChaseWiseman:
@deivamagalhaes thanks for looking! That's the JS error that this PR fixes so I'm not sure what might be wrong.

Do you have the `sv_wc_payment_gateway_payment_form_params` object printed in your footer? When you see the CSC message, is it after a page refresh, or does the JS scroll you to the top?

wvega:
@ChaseWiseman @deivamagalhaes the `composer.lock` file in https://github.com/skyverge/wc-plugins/pull/3717 has `dev-master` as the Framework version, so `composer install` doesn't download these changes.

I can see the same error using that branch as-is: https://cloud.skyver.ge/OAubdPw4

After `composer update`, the error is gone and the payment form works fine.

PR #497, opened 2020-06-02
Prevent jQuery conflict with Disable Admin Notices plugin

wvega:
# Summary

This PR fixes a JavaScript conflict between the framework's admin notice handler and the Disable Admin Notices plugin.

## Details

This is one is caused by a series of unfortunate events:

The JavaScript we see in the console is the result of an exception triggered in:

https://github.com/skyverge/wc-plugin-framework/blob/a9d6777e95377ce47c9d4d0f84f677c1c795d569/woocommerce/class-sv-wc-admin-notice-handler.php#L299

The exception occurs because the placeholder is being rendered inside one of the notices:

We render the notice and the placeholder correctly, but the Disable Admin Notices plugin [runs the callbacks attached to the `admin_notices` hook](https://plugins.trac.wordpress.org/browser/disable-admin-notices/trunk/includes/classes/class-configurate-notices.php?rev=2138539#L246) on its own and uses regular expressions to [inject](https://plugins.trac.wordpress.org/browser/disable-admin-notices/trunk/includes/classes/class-configurate-notices.php?rev=2138539#L294) the **Hide notification forever** link.

Unfortunately, the regular expression assumes that only one notice (or top level element) is rendered per callback. If two or more notifications (or top level elements) are rendered by the same callback, the regular expression causes the second and subsequent elements to become children of the first element.

In the notice handler case, the regular expression causes the placeholder to become a child of one of the notifications. That triggers an exception when we execute that JS line from above trying to move notifications after the placeholder.

Normally, that line shouldn't trigger an exception because the placeholder is rendered next to a visible notice, and the script tries to move hidden notices. However, WooCommerce has been rendering notices inside `<div class="woocommerce-layout__notice-list-hide" id="wp__notice-list"></div>` since v4.0. That causes our visible notices to become hidden when the page loads, and as a result, that JS line tries to move the container of an elment, next to that very same element.

This PR uses JavaScript to check whether there are framework placeholders inside notifications and move them out.

The snippet targets placeholders rendered by the admin notice handler included in any plugin, not just the plugin that loaded this version of the framework. The goal is to prevent the JavaScript exception without having to role out an update on every plugin.

To achieve that goal, however, I had to render the snippet on every admin page and make sure it was rendered before any other JavaScript generated by the admin notice handlers.

---

The snippet prevents the exception, but doesn't get the "Hide notification forever" link working on all of our notices. The main reason for that is that due to the bug in Disable Admin Notices plugin, only one link is injected regardless of the number of notices that we render. As a result, the second and subsequent notices from the same plugin don't get the link.

### Setup

- Activate Elavon Converge using this branch of the framework: `dev-ch54179/conflict-with-disable-admin-notices-by-webcraftic` (easiest because there is no need to change the framework namespace)
- Activate Facebook for WooCommerce (don't change the framework branch)
- Add the following snippet to render an admin notice

```php
add_action( 'admin_init', function() {

	if ( function_exists( 'wc_elavon_converge' ) ) {
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 1!', 'wc-plugin-framework/ch54179-1' );
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 2!', 'wc-plugin-framework/ch54179-2' );
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 3!', 'wc-plugin-framework/ch54179-3' );
	}

	if ( function_exists( 'facebook_for_woocommerce' ) ) {
		facebook_for_woocommerce()->get_admin_notice_handler()->add_admin_notice( 'Hello Facebook 1!', 'wc-plugin-framework/ch54179-facebook-1' );
		facebook_for_woocommerce()->get_admin_notice_handler()->add_admin_notice( 'Hello Facebook 2!', 'wc-plugin-framework/ch54179-facebook-2' );
		facebook_for_woocommerce()->get_admin_notice_handler()->add_admin_notice( 'Hello Facebook 3!', 'wc-plugin-framework/ch54179-facebook-3' );
	}
} );
```

### Steps

1. Disable Elavon
1. Explore WooCommerce Orders, Edit Order, Products, Edit Products, and Settings pages to confirm there are JavaScript errors in the console
    - [x] The `Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The new child element contains the parent.` appears in the console
1. Enable Elavon
1. Explore WooCommerce Orders, Edit Order, Products, Edit Products, and Settings pages to confirm there are no JavaScript errors in the console
    - [x] There are no JavaScript errors in the console

### DIFF
--- a/tests/integration/Integrations/DisableAdminNoticesTest.php
+++ b/tests/integration/Integrations/DisableAdminNoticesTest.php
@@ -1,0 +1,24 @@
+<?php
+
+use SkyVerge\WooCommerce\PluginFramework\v5_7_1\Integrations\Disable_Admin_Notices;
+
+class DisableAdminNoticesTest extends \Codeception\TestCase\WPTestCase {
+
+
+	/** @var \IntegrationTester */
+	protected $tester;
+
+
+	/** Tests *********************************************************************************************************/
+
+
+	/** @see Integrations::get_integrations() */
+	public function test_constructor() {
+
+		$integration = new Disable_Admin_Notices();
+
+		$this->assertEquals( 10, has_action( 'admin_footer', [ $integration, 'enqueue_conflict_fix_script' ] ) );
+	}
+
+
+}
--- a/tests/integration/Integrations/IntegrationsTest.php
+++ b/tests/integration/Integrations/IntegrationsTest.php
@@ -1,0 +1,61 @@
+<?php
+
+use SkyVerge\WooCommerce\PluginFramework\v5_7_1\Integrations\Disable_Admin_Notices;
+use SkyVerge\WooCommerce\PluginFramework\v5_7_1\Integrations\Integrations;
+use SkyVerge\WooCommerce\PluginFramework\v5_7_1\SV_WC_Plugin;
+
+class IntegrationsTest extends \Codeception\TestCase\WPTestCase {
+
+
+	/** @var \IntegrationTester */
+	protected $tester;
+
+	/** @var Integrations */
+	protected $integrations;
+
+
+	protected function _before() {
+
+		// include plugins with integrations as active
+		// wp-browser also adds a handler for pre_option_active_plugins to return the plugins in Codeception's configuration
+		// See wp_tests_options() in vendor/lucatume/wp-browser/src/includes/bootstrap.php
+		add_filter( 'pre_option_active_plugins', function( $plugins ) {
+
+			if ( ! in_array( 'disable-admin-notices/disable-admin-notices.php', $plugins, true ) ) {
+				$plugins[] = 'disable-admin-notices/disable-admin-notices.php';
+			}
+
+			return $plugins;
+		} );
+
+		// remove active plugins cache
+		$property = new ReflectionProperty( SV_WC_Plugin::class, 'active_plugins' );
+		$property->setAccessible( true );
+		$property->setValue( sv_wc_test_plugin(), [] );
+
+		$this->integrations = new Integrations( sv_wc_test_plugin() );
+	}
+
+
+	/** Tests *********************************************************************************************************/
+
+
+	/** @see Integrations::get_integrations() */
+	public function test_get_integrations() {
+
+		$integrations = $this->integrations->get_integrations();
+
+		$this->assertArrayHasKey( Integrations::INTEGRATION_DISABLE_ADMIN_NOTICES, $integrations );
+	}
+
+
+	/** @see Integrations\Integrations::get_integration() */
+	public function test_get_integration() {
+
+		$integration = $this->integrations->get_integration( Integrations::INTEGRATION_DISABLE_ADMIN_NOTICES );
+
+		$this->assertInstanceOf( Disable_Admin_Notices::class, $integration );
+	}
+
+
+}
--- a/woocommerce/Integrations/Disable_Admin_Notices.php
+++ b/woocommerce/Integrations/Disable_Admin_Notices.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * WooCommerce Plugin Framework
+ *
+ * This source file is subject to the GNU General Public License v3.0
+ * that is bundled with this package in the file license.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://www.gnu.org/licenses/gpl-3.0.html
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@skyverge.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade the plugin to newer
+ * versions in the future. If you wish to customize the plugin for your
+ * needs please refer to http://www.skyverge.com
+ *
+ * @package   SkyVerge/WooCommerce/Plugin/Classes
+ * @author    SkyVerge
+ * @copyright Copyright (c) 2013-2020, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_7_1\Integrations;
+
+defined( 'ABSPATH' ) or exit;
+
+if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_7_1\\Integrations\\Disable_Admin_Notices' ) ) :
+
+
+/**
+ * Disable Admin Notices Integration
+ *
+ * @link https://wordpress.org/plugins/disable-admin-notices/
+ *
+ * @since 5.7.2-dev.1
+ */
+class Disable_Admin_Notices {
+
+
+	/**
+	 * Constructor.
+	 *
+	 * @since 5.7.2-dev.1
+	 */
+	public function __construct() {
+
+		$this->add_hooks();
+	}
+
+
+	/**
+	 * Setup integration hooks.
+	 *
+	 * @since 5.7.2-dev.1
+	 */
+	protected function add_hooks() {
+
+		// priority must be strictly less than 20 - see callback description
+		add_action( 'admin_footer',  [ $this, 'enqueue_conflict_fix_script' ], 10 );
+	}
+
+
+	/**
+	 * Enqueues a JavaScript snippet used to prevent an Uncaught DOMException when Disable Admin Notices is active.
+	 *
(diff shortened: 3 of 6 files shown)


wvega:
@deivamagalhaes thanks for the review. I see a floating link too on the Plugins page when the same notice handler prints two or more notices.

The source of the problem is the same. This [call](https://plugins.trac.wordpress.org/browser/disable-admin-notices/trunk/includes/classes/class-configurate-notices.php?rev=2138539#L294) to `preg_replace()` in the Disable Admin Notices plugin assumes only one notice is included in the content and inevitably breaks anything that includes two notices or more.

I think we'll have to use the JavaScript approach after all.

Could you please try using `dev-ch54179/another-approach` as the framework version?

That branch includes this commit: https://github.com/skyverge/wc-plugin-framework/commit/b92e148520c555f99ca71736e963813520514ad7. This is what I see using the following snippet:

```php
add_action( 'admin_init', function() {

	if ( function_exists( 'wc_elavon_converge' ) ) {
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 1!', 'wc-plugin-framework/ch54179-1' );
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 2!', 'wc-plugin-framework/ch54179-2' );
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 3!', 'wc-plugin-framework/ch54179-3' );
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 4!', 'wc-plugin-framework/ch54179-4' );
		wc_elavon_converge()->get_admin_notice_handler()->add_admin_notice( 'Hello World 5!', 'wc-plugin-framework/ch54179-5' );
	}
} );
```

* Products: https://cloud.skyver.ge/qGuKzjmj
* Edit Product: https://cloud.skyver.ge/lluJwjKJ
* Orders: https://cloud.skyver.ge/8Lu7ZBOq
* Edit Orders: https://cloud.skyver.ge/WnubA2g5
* Settings: https://cloud.skyver.ge/eDu1LQdg
* Plugins: https://cloud.skyver.ge/YEupvgnN

deivamagalhaes:
@wvega , using that branch of the framework, I see the same as you, I do see my notices and I do not see floating Hide links, but why is there only one Hide link for multiple notices by the same plugin?

wvega:
@deivamagalhaes that's because Disable Admin Notices doesn't handle multiple notices printed by the same callback very well.

In this case, the notice handler has to render five or six notices, so it generates the following HTML content:

```html
<div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
    data-message-id="wc-plugin-framework/ch54179-1">
    <p>Hello World 1!</p>
</div>
<div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
    data-message-id="wc-plugin-framework/ch54179-2">
    <p>Hello World 2!</p>
</div>
<div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
    data-message-id="wc-plugin-framework/ch54179-3">
    <p>Hello World 3!</p>
</div>
<div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
    data-message-id="wc-plugin-framework/ch54179-4">
    <p>Hello World 4!</p>
</div>
<div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
    data-message-id="wc-plugin-framework/ch54179-5">
    <p>Hello World 5!</p>
</div>
<div class="notice js-wc-plugin-framework-admin-notice updated" data-plugin-id="elavon_vm"
    data-message-id="wc-elavon-welcome">
    <p><strong>WooCommerce Elavon Converge is almost ready!</strong> To get started, please ​<a
            href="https://plugins.skyverge.test/wp-admin/admin.php?page=wc-settings&#038;tab=checkout&#038;section=elavon_converge_credit_card">connect
            to Elavon Converge</a>.</p>
</div>
<div class="js-wc-elavon-vm-admin-notice-placeholder"></div>
```

Disable Admin Notices then uses preg_replace [here](https://plugins.trac.wordpress.org/browser/disable-admin-notices/trunk/includes/classes/class-configurate-notices.php?rev=2138539#L294) to insert the Hide link. However, as you can see from the regular expression (`/^(<div[^>]+>)(.*?)(<\/div>)$/`), it assumes there is a single notice.

When the plugin insert the links, the HTML content is transformed into:

```
<div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
    data-message-id="wc-plugin-framework/ch54179-1">
    <div class="wbcr-dan-hide-notices">
        <p>Hello World 1!</p>
    </div>
    <div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
        data-message-id="wc-plugin-framework/ch54179-2">
        <p>Hello World 2!</p>
    </div>
    <div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
        data-message-id="wc-plugin-framework/ch54179-3">
        <p>Hello World 3!</p>
    </div>
    <div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
        data-message-id="wc-plugin-framework/ch54179-4">
        <p>Hello World 4!</p>
    </div>
    <div class="notice js-wc-plugin-framework-admin-notice updated is-dismissible" data-plugin-id="elavon_vm"
        data-message-id="wc-plugin-framework/ch54179-5">
        <p>Hello World 5!</p>
    </div>
    <div class="notice js-wc-plugin-framework-admin-notice updated" data-plugin-id="elavon_vm"
        data-message-id="wc-elavon-welcome">
        <p><strong>WooCommerce Elavon Converge is almost ready!</strong> To get started, please ​<a
                href="https://plugins.skyverge.test/wp-admin/admin.php?page=wc-settings&#038;tab=checkout&#038;section=elavon_converge_credit_card">connect
                to Elavon Converge</a>.</p>
    </div>
    <div class="js-wc-elavon-vm-admin-notice-placeholder"><a href="#" data-nonce="1f3f173ccd"
            data-notice-id="f21598d37f911d432eeb2c6f09a3126f_aa84826fcc8abb4ada37630566d8d150"
            class="wbcr-dan-hide-notice-link">[Hide notification forever]</a></div>
</div>
```

Now all but the first notice and the placeholder are inside the first notice. Since the plugin doesn't handle multiple notices, it inserted a single link at the end (which ended up being inside the placeholder). That's the single link you see, after the [script in that framework branch](https://github.com/skyverge/wc-plugin-framework/commit/b92e148520c555f99ca71736e963813520514ad7) tries to put it in the right place.

I don't think there is a way for us to make those links work. The best we can hope for is to prevent the JavaScript error mentioned in the story. However, on that regard, I noticed that this fix works only if all plugins using the framework update to this version. If we go the JavaScript route, perhaps we should use the script to fix notices generated by any copy of the framework, otherwise is going to take a while to deploy the fix to all the plugins.

wvega:
@deivamagalhaes thank you for the early review and testing the two approaches in this PR. I'll request a team review for the latest changes.

I replaced the previous commits with the changes for the JavaScript approach and updated the snippet to prevent the error that could be triggered by the JavaScript code in other versions of the framework.

wvega:
@deivamagalhaes thank you for those suggestions. The implementation using an integration class looks a lot cleaner.

We didn't have integrations at the plugin level, so I added an `Integrations` class to check whether the Disable Admin Notices plugin is active and instantiate an integration for that plugin if necessary.

wvega:
Thank you @deivamagalhaes. That's right, since they process all the notifications rendered in the same callback as one, clicking the link causes Disable Admin Notices to hide all the notifications.

I don't know how to fix it on our end, or prevent that without causing other problems.

They [use `md5` to create a unique ID for the notification](https://plugins.trac.wordpress.org/browser/disable-admin-notices/trunk/includes/classes/class-configurate-notices.php?rev=2138539#L254), so a small modification in the content should bring the notices back, because it would appear to the plugin as a completely different notification. However, that's not really fix and I don't think we should introduce something like that.

Once we approve and merge this, I plan to post a support request on the forums explaining the problem and what we are doing. I think the best place to properly fix these issues is on their plugin by adding support for multiple notices rendered in the same callback.

wvega:
Thanks @deivamagalhaes. I started a support conversation in https://wordpress.org/support/topic/javascript-conflict-with-skyverge-plugins/

PR #500, opened 2020-07-31
Release: 5.8.1

unfulvio:
# Summary

* Bumps FW to 5.8.1 / namespaces to 5_8_1
* Ensures there are no scripts with namespaces pointing to 5_7_1

## Before merge

- [x] Integration tests pass
- [x] All required plugin changes have been documented for this new version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
     "name": "skyverge/wc-plugin-framework",
     "description": "The official SkyVerge WooCommerce plugin framework",
-    "version": "5.8.0",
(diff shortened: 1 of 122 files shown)


wvega:
Added a small commit to update some very outdated namespaces in archived tests. We are not using those, but it can become confusing if we have namespaces that need to update, and namespaces that we can ignore: https://github.com/skyverge/wc-plugin-framework/pull/500/commits/1077f03c3dd194ea79dd65a0837f951eb20fd90c

PR #502, opened 2020-09-28
Add `SV_WC_Plugin_Compatibility::is_rest_api_request()` 

unfulvio:
## Summary

Adds an helper method which is only available in recent WC versions `\WooCommerce::is_rest_api_request()` to `SV_WC_Plugin_Compatibility::is_rest_api_request()`. The method detects whether the current is a request for a WordPress REST API route.

## Additional details

For an example usage of a plugin that implements a similar method, see Memberships 1.19.1: https://github.com/skyverge/woocommerce-memberships/pull/933

WC Subscriptions has a similar implementation too. 

## QA

- [ ] Code review

### DIFF
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -331,7 +331,7 @@ class SV_WC_Plugin_Compatibility {
 	 *
 	 * @see \WooCommerce::is_rest_api_request()
 	 *
-	 * @since 5.7.2-dev.1
+	 * @since x.y.z-dev.1
 	 *
 	 * @return bool
 	 */
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -326,33 +326,6 @@ class SV_WC_Plugin_Compatibility {
 	}
 
 
-	/**
-	 * Determines if the current request is for a WC REST API endpoint.
-	 *
-	 * @see \WooCommerce::is_rest_api_request()
-	 *
-	 * @since x.y.z-dev.1
-	 *
-	 * @return bool
-	 */
-	public static function is_rest_api_request() {
-
-		if ( is_callable( 'WC' ) && is_callable( [ WC(), 'is_rest_api_request' ] ) ) {
-			return (bool) WC()->is_rest_api_request();
-		}
-
-		if ( empty( $_SERVER['REQUEST_URI'] ) || ! function_exists( 'rest_get_url_prefix' ) ) {
-			return false;
-		}
-
-		$rest_prefix         = trailingslashit( rest_get_url_prefix() );
-		$is_rest_api_request = false !== strpos( $_SERVER['REQUEST_URI'], $rest_prefix );
-
-		/* applies WooCommerce core filter */
-		return (bool) apply_filters( 'woocommerce_is_rest_api_request', $is_rest_api_request );
-	}
-
-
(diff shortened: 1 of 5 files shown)


asseti6:
@unfulvio any reason not to put a unit test on this one?  Given that it is a boolean check and would be relied upon elsewhere.. would be good to make sure this doesn't have any holes and/or they are plugged as they go.

unfulvio:
@asseti6 this method copies 1:1 a function from WooCommerce which is not available in some older versions of WC which we support.

The method's body presumes some WP core functions and thus it needs to be an integration test.

I've added one here: https://github.com/skyverge/wc-plugin-framework/pull/502/commits/f6f0cec14e00dc2ce387510cbc948b698badb94d however, I am not sure how to override the `$_SERVER` variable in a way that the function returns true. I have tried by just setting the key the function expects but it never passes. The WP functions I need are confirmed to be available. 

How to mock that server request in an integration test context?

woocommerce/class-sv-wc-plugin-compatibility.php
```diff
@@ -326,6 +326,33 @@ public static function is_enhanced_admin_available() {
...
+	 *
+	 * @see \WooCommerce::is_rest_api_request()
+	 *
+	 * @since x.y.z-dev.1
+	 *
+	 * @return bool
+	 */
+	public static function is_rest_api_request() {
+
+		if ( function_exists( 'WC' ) && is_callable( [ WC(), 'is_rest_api_request' ] ) ) {
```
wvega, line 340: Since we are already using in the second part of the condition I figured we could make the first part as strict:

```suggestion
		if ( is_callable( 'WC' ) && is_callable( [ WC(), 'is_rest_api_request' ] ) ) {
```

wvega (commented):
@unfulvio 

```php
		$_SERVER['REQUEST_URI'] = '/wp-json/';

**Review by wvega — COMMENTED** [review]
@unfulvio I'm sorry about the previous comment. I submitted it way before it was ready. Here is what I meant to post:

>I have tried by just setting the key the function expects but it never passes.

Adding the following lines at the top of the test method makes the current test fail for me (i.e. make `Framework\SV_WC_Plugin_Compatibility::is_rest_api_request()` return `true`). Is that what you tried?

```
$_SERVER['REQUEST_URI'] = '/wp-json/';
```

Other than that I think is hard to test the part of the code that we copied from WooCommerce, because `WC()` is always going to be available on Integration tests and the method uses too much of WordPress to use a unit test.

Code looks good to me. If the line above works for you, maybe we can add a `@dataProvider` to test a positive and a negative case. Otherwise I would say we  can merge.

unfulvio:
oh @wvega thanks, I was declaring `global $_SERVER` first before overwriting its key, which seemed to make sense, but if I did that somehow the test didn't work - if I skip that and assign directly it does as you mentioned! I have added the data provider and the 2 possible test outcomes

unfulvio:
ps - by the way is this fine to merge to master... or? IIRC we agreed to merge these small changes to master, but are we still converting x.y.z to the actual release version?

wvega:
@unfulvio merging to master and keeping `x.y.z-dev.1` sounds good to me, but we should add a changelog entry to still keep track of the things we have added (added one in 8b02eb86).

We can replace `x.y.z-dev.1` once we decide to release another version.

If I want to use this new method I would have to update the framework to a version that includes it, but that's only possible if we tag a new release with it. So that could be another consideration. If we are going to update a plugin to take advantage of this method have to make a release, otherwise keeping it in master works until we get another reason to release.

PR #504, opened 2020-10-07
Address a PHP warning if no card logos are configured to be displayed at checkout

unfulvio:
## Summary

Hardens code so that a non-array is not passed to `array_map`.

## Details

There are already tests for the functions involved, I'm not sure how to make `get_js_handler_args()` return the additional array key that would make use of the piece involved in this fix without modifying the mock plugin we use to run our integration tests. 

## QA

- [x] Code review


N-Molham:
@unfulvio code looking good

PR #505, opened 2020-10-07
Fix a Capture Charge button issue when adding items to an order

adrianostanley:
# Summary

This PR will fix an issue with the _Capture Charge_ button after adding an item to an existing order.

pMpG1BT4Lw

## Details

When an order page is loaded, the `html-order-partial-capture.php` is added to the page and the `sv-wc-payment-gateway-admin-order.coffee` script attaches events to the _Capture Charge_ button. This works well when the page loads.

However, when adding an item (like the fee added in the recorded above), the same template is called and the contents of that box are changed. The script, however, is not called anymore since it's attached to the body ready event.

To fix that, this PR will:

* Move the events attachments to a `window` function called `sv_wc_payment_gateway_admin_order_add_capture_events`;
* Call this function when the page is ready to keep the original behavior
* Call this function in the template to reattach all the events
    * **Note:** maybe adding scripts to template files are not a good practice but I couldn't think of a quick solution so feel free to request changes with other approaches.

**Note to the reviewer:** ignore commit https://github.com/skyverge/wc-plugin-framework/commit/07353953c0b8e6a45b8eda8601e5a90a15762b05 as it was for an approach I discarded.

## QA

The original bug is represented in the description as a GIF. But feel free to branch `master` and see it happening. By following the QA steps you may be able to reproduce it.

### Setup

1. Activate Bambora or Global Payments plugin
2. Open the `composer.json` and change replace the framework version
    * `"skyverge/wc-plugin-framework": "dev-ch57945/capture-charge-button-after-adding-fee-hides"`
3. Do a `composer update skyverge/wc-plugin-framework`
4. Set to "Authorization" mode and allow partial capture
5. Create an order using one of these gateways

### Steps

1. Go to WooCommerce > Orders
2. Click on the order you created in the setup process
3. Scroll down and click on the _Capture Charge_
    - [x] The capture charge box slides and gets visible
4. Click on Cancel
5. Click on _Add item(s)_
6. Click on _Add fee_ (you may also want to add products, shipping, etc)
7. Type 1 in the fee popup
8. Wait until the fee is added and the box is reconstructed
9. Click on _Capture Charge_
    - [x] The capture charge box slides and gets visible

## Before merge


N-Molham (approved):
👌👌

PR #508, opened 2020-10-15
Add Visa Checkout settings

wvega:
# Summary

This PR updates the framework to define the Visa Checkout feature and show a basic set of settings for plugins that support it.

## Details

I branched of version v5.8.1 so that it's easy to test these changes on plugins without having to update the namespaces. We will likely have to rebase the changes here and integrate them with the Google Pay modifications before we start thinking about merging this one into a new framework version. So, for now, the target version is `feature/visa-checkout` which is the same as the `5.8.1` tag.

The Visa Checkout settings borrow almost everything from Apple Pay. This PR doesn't modify the existing Apple Pay handlers, but I chose to copy the different methods and classes into base handlers and handlers specific to Visa Checkout. 

I tried to put some of the logic on a trait, but I couldn't come with a set of methods that we would want to re-use for something other than creating settings for providers such as Apple Pay, Google Pay, and Visa Checkout. **Maybe the entire base handler is a trait?**

Perhaps the one I wanted to move the most was `Settings_Screen:: render_static_setting()` because I would rather have the `static` setting type globally available when our plugins are loaded, instead of having each class that uses settings and needs something like that define a custom type that does the same (we have something similar on Facebook settings). However, I don't think we have a place where we can register new field types globally and I chose not to introduce one here.

This PR introduces a `SkyVerge\WooCommerce\PluginFramework\v5_8_1\Payment_Gateway\Settings_Screen` that provides the common functionality to add, render, and save a settings section such as the Apple Pay settings section.

The `SkyVerge\WooCommerce\PluginFramework\v5_8_1\Payment_Gateway\Visa_Checkout\Admin` extends `Settings_Screens` to define the new **Visa Checkout** settings section.

This PR also introduces `SkyVerge\WooCommerce\PluginFramework\v5_8_1\Payment_Gateway\Visa_Checkout\Visa_Checkout` the main handler for the Visa Checkout feature. This one also copies a lot from the corresponding handler for Apple Pay and I think it makes sense to split some of that into a base handler that we can use for Visa Checkout and Google Pay. Here I just copied the parts that I ended to setup the settings section.

**What would be a good name for a common handler for Google Pay, Apple Pay, Visa Checkout, and others like them?** 

CyberSource docs kind of refer to them as *Online Authorizations*, *Payment Network Tokenization Providers*, or *Digital Payment Solutions*.

## UI Changes

- New Visa Checkout settings: https://cloud.skyver.ge/qGuvXYoz

### Setup

1. Install the `TBD` branch of CyberSource
1. Add the following snippet to activate Visa Checkout:

        add_filter( 'wc_payment_gateway_cybersource_activate_visa_checkout', '__return_true' );

#### Settings

1. Go to WooCommerce > Settings > Payments > Visa Checkout
    - [ ] The Enable/Disable Visa Checkout setting is shown
    - [ ] The Environment select is shown
    - [ ] The API Key setting is shown
    - [ ] The Processing Gateway static setting is shown
1. Enable Visa Checkout
1. Select a different environment
1. Enter a value in the API Key field
1. Save changes
    - [ ] The fields are saved 

#### Tests 

- [ ] `codecept run acceptance` pass

## Before merge

### DIFF
--- a/tests/_support/AcceptanceTester.php
+++ b/tests/_support/AcceptanceTester.php
@@ -16,11 +16,82 @@
  *
  * @SuppressWarnings(PHPMD)
 */
-class AcceptanceTester extends \Codeception\Actor
-{
+class AcceptanceTester extends \Codeception\Actor {
+
     use _generated\AcceptanceTesterActions;
 
-   /**
-    * Define custom actions here
-    */
+
+	/** Visa Checkout settings screen *****************************************/
+
+
+	/**
+	 * @since 5.10.0-dev.1
+	 */
+	public function haveVisaCheckoutActivated( $plugin_id ) {
+
+		$this->haveMuPlugin(
+			'activate-visa-checkout.php',
+			"add_filter( 'wc_payment_gateway_{$plugin_id}_activate_visa_checkout', '__return_true' );"
+		);
+	}
+
+
+	/**
+	 * @since 5.10.0-dev.1
+	 */
+	public function amOnVisaCheckoutSettingsPage() {
+
+		$this->amOnAdminPage( 'admin.php?page=wc-settings&tab=checkout&section=visa-checkout' );
+	}
+
+
+	/**
+	 * @since 5.10.0-dev.1
+	 */
+	public function checkVisaCheckoutEnableSetting() {
+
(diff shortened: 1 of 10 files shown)


wvega:
Closing. The changes added as part of the Google Pay epic supersede the modifications added here.

PR #509, opened 2020-10-16
Add parcel to build 

alexanderchan:
# Summary

This PR updates the build commands to use Parcel for the payment gateway JS files.

## Details

This PR also updates the PHP to enqueue the files from the new folder (inside a `dist` folder and without the `.min` suffix) and removes the now unused Babel and Uglify Grunt tasks and their dependencies.

### Future improvements

- Build the JS files outside of `woocommerce/payment-gateway`
- Transpile coffee to .js and remove .coffee files

## UI Changes

N/A

### Steps

1. Run `npm run build` to build the Framework files
    - [x] The payment gateway compiled and minified JS files for both admin and frontend are generated in the corresponding dist folder

## Before merge


alexanderchan:
@deivamagalhaes this is just the minimum to build and dev the js/coffee for `woocommerce/payment-gateway/assets/js/frontend` I can do a bit more to clean up the other build since we don't need to uglify any more for these as well but I can do that in another PR.

alexanderchan:
@deivamagalhaes normally we would group all the compiled assets together into a `dist` folder but would that work for the plugin framework?

```
parcel build 'woocommerce/**/*.scss' 'woocommerce/**/*.js' 'woocommerce/**/*.coffee' --out-dir woocommerce/dist
```
could output everything under the woocommerce directory into the `woocommerce/dist` folder.  Could that also work or do the js assets need to live below the `payment-gateway` directory in order to be able to use them?  Otherwise we'll probably just have a build for each of the subfolders that need to have a dist below them.

deivamagalhaes:
@skyverge/engineering-wordpress, since Alex's question impacts all the minified assets on the Framework, I would like to know everybody's thoughts.

I'd personally prefer to keep the directories as they are, to avoid changes to other parts of the code.

alexanderchan:
We should still output to a different folder, it's best to split out built assets from original assets, and we have to do it for `*.js` files that are transpiled down to es5.  So I guess the question is, where are the `dist` folder(s) going to live?  

1) per area?
- `woocommerce/payment-gateway/assets/dist`
2) top level
- `woocommerce/dist` or `woocommerce/assets/dist`

It also helps us from picking up `coffee=>.js` assets and re-compiling them.

wvega:
@alexanderchan @deivamagalhaes I think adding the `dist` under each directory is better. Introducing that dir already requires changes to the framework, but I think is easier to keep them separated. 

I also believe `saké` removes the gateways part of the framework from non-gateway plugins, so that would be another reason to keep those separated.

Question: even if is not the best practice, is it possible to keep the files on the same directories they are now? As @deivamagalhaes mentioned, the update is going to be much easier if we don't have to change locations.

alexanderchan:
@wvega @deivamagalhaes it's harder to keep the same output folders especially when `.js` goes to `.js`.

I've just done it for the one folder for now with an example of the `.coffee` and `.js`  This will get us going for now and we can look at how to do it longer term with better wildcards.  `parcel` is good for not much configuration but gives us less options for input/output.

deivamagalhaes:
@alexanderchan, we discussed this during our WP Engineering meeting today and we were wondering if we could use the same approach used in [Sake](https://github.com/skyverge/sake/blob/master/pipes/scripts.js), to avoid having to add a dist folder. @wesleycole may also have additional information on that as I believe he was involved in the Sake implementation.

wesleycole:
As @alexanderchan recommended, it would be better to utilize a bundler like parcel instead of gulp.

1. Parcel will provide us with better longevity and a more feature-rich approach to compiling frontend assets. We can use parcel to compile JS (coffee, typescript, es5, etc.), CSS (Sass) or CSS with PostCSS for the transforms which will allow us to utilize new CSS features like custom properties (variables), import JSON into javascript files, and more, finally we will be able to use React automatically with Parcel. Utilizing babel with gulp will mean that any other future enhancements will have to be done separately where parcel takes care of all of that for us.
2. Having a `dist` folder is pretty commonplace now and is the same approach we took with the Dashboard. It will provide good separation between the src and dist files so that developers don't accidentally change the wrong file which could happen if they are kept together in the same directory as well as making it more difficult to compile when you have multiple `.js` files in the same directory.

Overall, using Parcel will provide us a more feature-rich and future friendly approach to bundling our frontend assets as well as an extremely minimal configuration. We can get rid of the vast majority of sake by just using `parcel` and running a simple `parcel-build` command like what @alexanderchan has in this PR. 

My vote is to make the changes now even if it is a little bit extra work because it will set us up for a better foundation for working with frontend assets in the future.

wesleycole:
Just to clarify, this is the directory that will be generated when parcel:

Screen Shot 2020-10-19 at 3 30 20 PM

Which means, that the majority of the work is changing the path when enqueuing the assets.

```
wp_enqueue_script( "$handle-v5_10_0", $this->get_plugin()->get_payment_gateway_framework_assets_url() . '/dist/frontend/' . $handle . '.min.js', array( 'jquery-tiptip', 'jquery' ), SV_WC_Plugin::VERSION );
```

The following: `$this->get_plugin()->get_payment_gateway_framework_assets_url()` will still return the same directory as before because we are outputting the dist files in `/assets` which is the same directory that the JS already exists in.

wesleycole:
The enqueue for the `woocommerce/admin` scripts will look like this:

```
wp_enqueue_script( 'sv-wc-admin-setup', $this->get_plugin()->get_framework_assets_url() . '/dist/admin/sv-wc-plugin-admin-setup-wizard.min.js', array( 'jquery', 'wc-enhanced-select', 'jquery-blockui' ), $this->get_plugin()->get_version() );
```

wesleycole:
@deivamagalhaes to answer your question about the approach in sake, that approach does not resolve the issue of polyfills. It might compile the es6 to vanilla, but you will still be missing some of the features like `Object.assign()` and Promises that we would have to add in addition to compiling via babel. 

This is another benefit of parcel, because it adds all of the appropriate polyfills for us in a single command.

deivamagalhaes:
Thanks for clarifying, @wesleycole.

@skyverge/engineering-wordpress, in light of this info, I agree it would be better to move forward with Parcel.

I can go ahead and make the necessary changes to this branch.

Any objections?

deivamagalhaes:
I've created a story for this: https://app.clubhouse.io/skyverge/story/66095/add-parcel-to-build-ecma-js-files

alexanderchan:
Thanks @deivamagalhaes I'll add on to that to write out some more details on that.

alexanderchan:
@deivamagalhaes this is looking pretty good, is there a reason we aren't putting the files in assets/dist like we had originally?  

Using a file glob like this will allow subfolders with js assets to be placed and mapped automatically, have everything processed in one parcel command, and make it easy to add on `*.scss`

```
parcel build 'woocommerce/payment-gateway/assets/js/**/*.js' 'woocommerce/payment-gateway/assets/js/**/*.coffee' --out-dir 'woocommerce/payment-gateway/dist/js'
```

then the path changes from:

```
/js/frontend/dist/sv-wc-payment-gateway-google-pay.js
to
/dist/js/frontend/sv-wc-payment-gateway-google-pay.js
```

deivamagalhaes:
@alexanderchan, as I mentioned on Slack I thought it made more sense to add the dist folder inside the js folder, since we are only changing the structure for js files. It is not a general assets dist folder, for now, but a js dist folder.

But if you have strong feelings about it I can change it back.

wesleycole:
@deivamagalhaes while it is just JS right now, you are making it a bit use case specific. There will come a time when we will need to use a bundler like parcel for all frontend assets (js, css, images, etc). By outputting everything to a general `assets/dist` directory instead of `assets/dist/js` we will have a lot more flexibility in the future. If we output to `/assets/js` we will have to make another PHP change to the plugins if we wanted to compile Sass/CS with parcel in the near future.

deivamagalhaes:
@alexanderchan and @wesleycole, I've changed so that the files from the `assets/js` folder are created in the `assets/dist` folder. The structure after the change is:
`assets/dist/admin/filename` and `assets/dist/frontend/filename`

alexanderchan:
> @alexanderchan and @wesleycole, I've changed so that the files from the `assets/js` folder are created in the `assets/dist` folder. The structure after the change is:
> `assets/dist/admin/filename` and `assets/dist/frontend/filename`
> At least we got to keep the separation between admin and frontend...
> 
> It can't be `assets/dist/js/admin/filename`, as mentioned by @alexanderchan on [this comment](https://github.com/skyverge/wc-plugin-framework/pull/509#issuecomment-712579615), because the source has to be limited to the `js` subfolder, otherwise, it includes the `dist` folder itself.

Ahh good point, I keep forgetting about the recursiveness, I mixed having a `src/` folder up because that would resolve that.  I'm ok with keeping that in mind but let's consider making the next revision where we consolidate assets in a different source folder as you mentioned @wvega had said.  

I'm good with this for now and I'll mention the alternative `src` folder as the longer term solution.

alexanderchan:
Here's the story [ch66189](https://app.clubhouse.io/skyverge/story/66189/migrate-grunt-to-parcel) for the longer term migration.

deivamagalhaes (approved):

PR #511, opened 2020-10-30
Add method to get IDs of gateways to inherit settings from

wvega:
# Summary

The framework currently doesn't support sharing settings between gateways when there are more than two gateways that share settings on the same plugin.

This adds a `SV_WC_Payment_Gateway::get_ids_of_gateways_to_inherit_settings_from()` helper method to allow gateways to control the IDs of gateways they can inherit settings from.

The method will be used in CyberSource to ensure that the eCheck and Visa Checkout inherit settings from the Credit Card gateway only.

## Details

This PR branches of `5.8.1` to facilitate testing on https://github.com/skyverge/woocommerce-gateway-cybersource/pull/13. However, the plan is to include this method the upcoming `v5.10.0`. That's why the since tag uses `5.10.0-dev.1`.

## QA

- [x] Code review only because user testing will be performed in https://github.com/skyverge/woocommerce-gateway-cybersource/pull/13

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -408,6 +408,19 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 
 
 	/**
+	 * Gets the IDs of sibling gateways that this gateway can inherit settings from.
+	 *
+	 * @since 5.10.0-dev.1
+	 *
+	 * @return array
+	 */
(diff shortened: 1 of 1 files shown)


unfulvio (approved):

wvega:
@unfulvio thank you for reviewing and merging! I'm going to restore the branch until https://github.com/skyverge/woocommerce-gateway-cybersource/pull/13 is reviewed. Composer has to use that branch instead of the release branch to avoid having to bump the framework namespace in that PR.

unfulvio:
sorry @wvega I got click happy here :)

PR #515, opened 2020-11-04
Add a sanity check to render_payment_field

adrianostanley:
# Summary

This PR adds a sanity check to prevent fields to be created if the name parameter is not informed.

## Details

Similar to #514: _branched of `5.8.1` tag so that we can use these changes on Elavon Converge without having to update the framework namespace._

### Steps

- [ ] Code review

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
@@ -977,10 +977,11 @@ class SV_WC_Payment_Gateway_Payment_Form extends Handlers\Script_Handler {
 	 */
 	protected function render_payment_field( $field ) {
 
-		if ( isset( $field['name'] ) ) {
-
-			woocommerce_form_field( $field['name'], $field, $field['value'] );
-		}
+		woocommerce_form_field(
+			isset ( $field['name'] ) ? $field['name'] : null,
+			$field,
+			isset ( $field['value'] ) ? $field['value'] : null
+		);
 	}
 
 
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
@@ -978,9 +978,9 @@ class SV_WC_Payment_Gateway_Payment_Form extends Handlers\Script_Handler {
 	protected function render_payment_field( $field ) {
 
 		woocommerce_form_field(
-			isset ( $field['name'] ) ? $field['name'] : null,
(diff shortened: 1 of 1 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -977,7 +977,10 @@ public function render_payment_fields() {
 	 */
 	protected function render_payment_field( $field ) {
 
-		woocommerce_form_field( $field['name'], $field, $field['value'] );
+		if ( isset( $field['name'] ) ) {
```
wvega, line 980: @adrianostanley I think this prevents the fields from https://github.com/skyverge/wc-plugins/pull/4121/files from showing up in the form. Wouldn't it?

I realize I didn't specify that on the document, but the intention is to remove the name attribute to prevent the browser from sending the value in those fields to the server. We still need those fields to get the values with JavaScript and pass them to Checkout.js

adrianostanley: @wvega oh, got it now, sorry.

Done in https://github.com/skyverge/wc-plugin-framework/pull/515/commits/9946f43b43dbf509714e6b0e7d45ce14050807e9.

It won't affect the other PR that was already merged, so we're good here.

Thank you for reviewing.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -977,7 +977,11 @@ public function render_payment_fields() {
 	 */
 	protected function render_payment_field( $field ) {
 
-		woocommerce_form_field( $field['name'], $field, $field['value'] );
+		woocommerce_form_field(
+			isset ( $field['name'] ) ? $field['name'] : null,
+			$field,
+			isset ( $field['value'] ) ? $field['value'] : null
```
wvega, line 983: Any particular reason you added a space after `asset()`?

```suggestion
			isset( $field['name'] ) ? $field['name'] : null,
			$field,
			isset( $field['value'] ) ? $field['value'] : null
```

wvega (approved):
Looks good @adrianostanley. Thank you for the updates!

All I have is a question/suggestion about style.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
adrianostanley: 😑 no particular reason... just lack of attention. Thanks for pointing that! Fixed in https://github.com/skyverge/wc-plugin-framework/pull/515/commits/ea2dcbbff107e1f854da76a57d5438d5555b4602.

PR #521, opened 2020-11-20
Remove admin_head action call for wizard steps

adrianostanley:
# Summary

Due to WordPress action calls workflow, `Setup_Wizard` was calling `<?php do_action( 'admin_head' ); ?>` before `current_screen` hook ([more details here](https://app.clubhouse.io/skyverge/story/67775/investigate-and-fix-php-notices-thrown-in-some-wizard-steps#activity-68207)).

Apparently, this action is not necessary as wizard steps don't include the admin header.

This PR will remove this action call because after PHP 8.0, what is now restricted to PHP Notices will become fatal errors and will prevent the wizard steps to be presented to users.

### Setup

1. Make sure you're running a PHP 8.0 environment

### Steps

1. Pick a plugin that has wizard steps, like Intuit (it will be used for the next steps, but it'd be good if this PR could also be validated with another plugin)
1. Activate Intuit
1. In the plugins page click on **Setup** below WooCommerce Intuit Payments Gateway then you'll be redirected to its wizard steps
1. A fatal error is thrown, preventing the first step to show
1. Open `composer.json` and bump `skyverge/wc-plugin-framework` version to `dev-ch67775/investigate-and-fix-php-notices-thrown-in`
1. Run `composer update skyverge/wc-plugin-framework`
1. You also may have to run a "replace all" from `5_8_1` to `5_10_0` to bump framework version and also disable other gateways and plugins that make use of the FW
1. Refresh the wizard steps page
    - [x] No PHP Notices nor errors are shown
    - [x] The wizard style remains the same
    - [x] The wizard scripts remain the same
    - [x] The wizard flow remains the same

## Before merge

### DIFF
--- a/woocommerce/admin/abstract-sv-wc-plugin-admin-setup-wizard.php
+++ b/woocommerce/admin/abstract-sv-wc-plugin-admin-setup-wizard.php
@@ -323,7 +323,6 @@ abstract class Setup_Wizard {
 				<?php wp_print_scripts( 'wc-setup' ); ?>
 				<?php do_action( 'admin_print_scripts' ); ?>
 				<?php do_action( 'admin_print_styles' ); ?>
-				<?php do_action( 'admin_head' ); ?>
 			</head>
 			<body class="wc-setup wp-core-ui <?php echo esc_attr( $this->get_slug() ); ?>">
 				<?php $this->render_header(); ?>
--- a/woocommerce/changelog.txt
+++ b/woocommerce/changelog.txt
@@ -2,6 +2,7 @@
 
 2020.nn.nn - version 5.10.1-dev.1
  * Fix - Prevent a deprecated notice for SV_WC_Payment_Gateway_Hosted::do_invalid_transaction_response() in PHP 8
+ * Fix - Prevent PHP Notices and fatal errors on wizard steps
 
 2020.11.06 - version 5.10.0
  * Feature - Add Google Pay support


adrianostanley:
**Note to reviewer:** While testing these changes, I realized a bug on the connect step on Intuit and by looking at the console, it looks to me this is an issue with my host (localhost) and since I can't run a `https://` instance on ngrok, I'm assuming that's causing the rendering issue. So, please let me know if you see the button.

image

I can see it on my Lando environment, running a `https` site and with the FW changes, but with PHP 7.

unfulvio (approved):
LGTM

unfulvio:
hey @adrianostanley I see you have also requested a review from @ChaseWiseman separately. This PR looks good to me and addresses the issue we saw in Setup Wizard implementations. Since you have requested the separate review from Chase I haven't merged in #519 - feel free to take action on this PR. Then, perhaps make a note/issue PR change on the plugins that implement a Setup Wizard so when 5.10.1 is released they can make use of the new FW version to fix the problem on their side. Thank you!

adrianostanley:
> Since you have requested the separate review from Chase I haven't merged in #519 .

@unfulvio I requested a review from you both separately since you were in the discussion in Slack, but I think we're good if you think the changes are ok.

> Then, perhaps make a note/issue PR change on the plugins that implement a Setup Wizard so when 5.10.1 is released they can make use of the new FW version to fix the problem on their side.

I'm sorry, but how this note is done? Still, is there any direct way to determine which plugins make use of Setup Wizard or should I enable all of them to check?

unfulvio:
> > Then, perhaps make a note/issue PR change on the plugins that implement a Setup Wizard so when  5.10.1 is released they can make use of the new FW version to fix the problem on their side.

> I'm sorry, but how this note is done? Still, is there any direct way to determine which plugins make use of Setup Wizard or should I enable all of them to check?

@adrianostanley sorry I did not list the current plugins currently implementing a Setup Wizard; these should be:

- Social Login
- Intuit 
- Memberships

I'll make a note to update to 5.10.1 as we will deploy the WC Jilt Promotions epic.

PR #522, opened 2020-12-01
Update deprecated jQuery usages

unfulvio:
## Summary

This PR updates some deprecated/legacy jQuery usages:

- Replaces jQuery `size()` with `length`
- Replaces `$.parseJSON` with `JSON.parse()`
- Updates jQuery document ready usages

Story: [CH 67466](https://app.clubhouse.io/skyverge/story/67466)

## Details

I've tested this branch while testing Global Payments PHP 8 & WC Jilt Promotions package removal PRs and couldn't spot problems. However, I'd like someone else in the team double checking before merging && tagging 5.10.2. Please check if you can spot also additional pieces we should update. Thanks!


woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-google-pay.js
```diff
@@ -1,4 +1,4 @@
-jQuery( document ).ready( ( $ ) => {
```
unfulvio, line 1: is there a specific reason for this arrow function here?

woocommerce/class-sv-wc-admin-notice-handler.php
```diff
@@ -254,53 +254,56 @@ public function render_admin_notice_js() {
 		self::$admin_notice_js_rendered = true;
 
 		ob_start();
+
 		?>
+		( function( $ ) {
```
unfulvio, line 259: @wvega pointed out that it may be best to wrap functions that end up in `wc_enqueue_js()` with an anonymous function call so that any variable set by these snippets in the global namespace do not get accidentally overwritten or overwrite third party code, since everything is funnelled into `wc_enqueue_js()` with a long string. So I've updated our usage here.

N-Molham (approved):

PR #523, opened 2020-12-07
Release 5.10.3

deivamagalhaes:
## Summary

Fix enqueuing the token editor JS (bug introduced in `5.10.0`).

### Stories
- [x] [CH 70180](https://app.clubhouse.io/skyverge/story/70180/token-editor-is-not-working)

### Before merge
- [x] Tests are passing
- [ ] The version is finalized and all namespaces have been updated


unfulvio (approved):
hey @deivamagalhaes this looks ok to me - and it seems it was only that script that needed to be enqueued unminified, the rest from `/dist` still used their unminified version. 

I could only catch a CoffeeScript JSDoc reference to the previous version number which I patched here: https://github.com/skyverge/wc-plugin-framework/pull/523/commits/b926276b99eaedb53c41b03c0b2aea8b59a7c7a1

The rest looks ok and there are no more references to 5_10_2.

This looks good to go to me. I see though you've added @ChaseWiseman on top of @skyverge/engineering-wordpress for review - does this need Chase's approval for release?

PS Performance/optimization related question outside the scope of this update: is there any reason why we don't minify those scripts?

dmagalhaes-godaddy:
> I could only catch a CoffeeScript JSDoc reference to the previous version number which I patched here: [b926276](https://github.com/skyverge/wc-plugin-framework/commit/b926276b99eaedb53c41b03c0b2aea8b59a7c7a1)

Thanks for catching that!

> This looks good to go to me. I see though you've added @ChaseWiseman on top of @skyverge/engineering-wordpress for review - does this need Chase's approval for release?

Not really, it should be good to go. I am not 100% clear on the "release" process though. Is it just merging and tagging?

> PS Performance/optimization related question outside the scope of this update: is there any reason why we don't minify those scripts?

We actually are minifying the scripts, they just don't have the `.min` suffix on the filename anymore, but all scripts on the `dist` folder are minified.

PR #524, opened 2021-01-11
Enable detailed messages by default

unfulvio-godaddy:
## Summary

Enables "Detailed declined messages" by default for gateways that support this feature.

Only affects fresh installations when setting a default value with initial settings. 

## QA

- [x] Code review


wvega (approved):
@unfulvio changes look good. LGTM!

PR #526, opened 2021-02-09
Add sanity checks to SV_WC_Payment_Gateway::payment_fields()

adrianostanley:
# Summary

This PR will add sanity checks in `SV_WC_Payment_Gateway::payment_fields()` to prevent incompatibility with page builders like Divi Builder and Elementor.

## Details

Some page builders like Divi Builder and Elementor may allow users to edit the checkout page. And while editing, the page may be rendered without any items in the cart, causing `$this->get_payment_form_instance()->render();` to be called while the payment form is `null`.

## QA

- [x] Code review

## Before merge

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -777,7 +777,7 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 	 */
 	public function payment_fields() {
 
-		if ( $this->supports_payment_form() ) {
+		if ( $this->get_payment_form_instance() && is_callable( [ $this->get_payment_form_instance(), 'render' ] ) && $this->supports_payment_form() ) {
 
 			$this->get_payment_form_instance()->render();
 


wvega (approved):
LGTM! I was able to reproduce the Fatal error using the steps below and confirmed that this branch fixes it.

- Activate both Authorize.Net and Elementor Pro
- Log in as admin
- Add a product to the cart
- Add `add_filter( 'woocommerce_checkout_redirect_empty_cart', '__return_false' );`
- Edit the checkout page with Elementor
- _Result_: Fatal error is thrown

Disabling the redirect "fixes" the "The preview could not be loaded" Elementor error shown while trying to edit the Checkout page.

The payment form doesn't show up, of course, but I believe that's fine while editing the page. Elementor doesn't allow to interact with any of the other Checkout fields anyway, so I believe there is no harm.

acastro1-godaddy:
> LGTM! I was able to reproduce the Fatal error using the steps below and confirmed that this branch fixes it.
> 
> * Activate both Authorize.Net and Elementor Pro
> * Log in as admin
> * Add a product to the cart
> * Add `add_filter( 'woocommerce_checkout_redirect_empty_cart', '__return_false' );`
> * Edit the checkout page with Elementor
> * _Result_: Fatal error is thrown
> 
> Disabling the redirect "fixes" the "The preview could not be loaded" Elementor error shown while trying to edit the Checkout page.
> 
> The payment form doesn't show up, of course, but I believe that's fine while editing the page. Elementor doesn't allow to interact with any of the other Checkout fields anyway, so I believe there is no harm.

@wvega good idea to use the filter! I had tried the same steps without it but could catch only the Elementor bug itself. Happy to see that this branch fixes it!

PR #527, opened 2021-03-04
Use the right format specifier in sprintf

unfulvio-godaddy:
## Summary

Fixes an issue with `sprintf()` in the external checkout admin handler as twi of the placeholders were `$S` (uppercase) and thus invalid. 

While fixing this I figured we didn't need as many placeholders for the HTML `<strong>` tags as these could be part of the same string, hardcoded and untranslatable (see diff).

## QA

- [x] Code review


wvega-godaddy (approved):
@unfulvio thank you for adding these changes. Code looks good!

We are going to have to make another framework release and update Braintree to use it, so I think we will be able to include this fix in 5.10.5.

PR #530, opened 2021-03-04
Check if array key exist in token editor template

unfulvio-godaddy:
## Summary 

While testing https://jira.godaddy.com/browse/MWC-263 I bumped in a PHP warning " Undefined array key "type" " in https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/payment-gateway/admin/views/html-user-payment-token-editor-token.php#L32. Perhaps this happened because I only had PayPal tokens and no credit card tokens? Nonetheless, we do have an `isset()` check for the options key, and one such check should also exist for the $field type key.

## QA

To replicate, I happened to follow the QA instructions in https://github.com/woocommerce/woocommerce-gateway-paypal-powered-by-braintree/pull/342 while I did not have any credit card tokens saved. 

- [x] Code review

## Before merge

- [ ] This PR makes the appropriate modifications to automated tests or explains why none are needed
- [ ] This PR makes the appropriate modifications to documentation or explains why none are needed
- [x] This change does not impact usage or expected outputs of covered code

### DIFF
--- a/woocommerce/payment-gateway/admin/views/html-user-payment-token-editor-token.php
+++ b/woocommerce/payment-gateway/admin/views/html-user-payment-token-editor-token.php
@@ -29,7 +29,7 @@
 
 	<?php foreach ( $fields as $field_id => $field ) : ?>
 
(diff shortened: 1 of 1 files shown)


adrianostanley (approved):
LGTM!

PR #531, opened 2021-03-10
Payment form JS is never loaded at checkout if a $0 order has a total added later [6]

N-Molham:
## Summary

This PR solves the issue where payment fields won't initialize when the customer starts the checkout with $0 order.

## QA

1. Configure the Cybersource gateway and confirm the payment form JS is functional (any other direct gateway you have running works as well)
2. Add a shippable product that costs $0
3. Add a shipping zone that applies to a specific region or zip code, and a shipping method to it that costs $1
4. Ensure the fallback "Locations not covered by your other zones" zone has no methods enabled
5. Visit the product in an incognito window
6. Add to cart and visit checkout
    - The total is $0 (no shipping chosen yet)
    - No payment form is present
7. Fill in your billing details so that it matches your shipping zone and shipping becomes available
    - The order is > $0
    - [ ] The payment form is loaded, and once the method selected the JS get applied (auto-formatting, etc...)


woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
```diff
@@ -64,6 +64,12 @@ jQuery ( $ ) ->
 		# Returns nothing.
 		handle_checkout_page: ->
 
+			# make sure payment fields initialized upon method selection
+			$( document.body ).on( 'payment_method_selected', =>
```
ChaseWiseman, line 68: what prompted this change? The form JS should have already been loaded, via the `updated_checkout` even that calls `this.set_payment_fields` below.

If I revert this change (and fix the issues from my other comments) then the form seems to load okay. Was there a case where you saw it not working?

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -99,7 +99,22 @@ protected function add_hooks() {
...
+	 * Loads payment gateway render script if checkout page or pay page.
+	 *
+	 * @since 5.10.5
+	 */
+	public function load_render_js() {
+
+		// also make sure it's not order received page
+		if ( ! is_order_received_page() && ( is_checkout() || is_checkout_pay_page() ) ) {
+			// payment form JS
+			add_action( 'wp_footer', [ $this, 'render_js' ] );
```
ChaseWiseman, line 116: This is being called when already inside the `wp_footer` action, so it doesn't end up rendering the JS.

Rather than the `load_render_js` wrapping function, we should just use `add_action( 'wp_footer', [ $this, 'render_js' ] );` in `add_hooks()`, then in `render_js` add the logic to bail early if not on the proper pages

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -99,7 +99,22 @@ protected function add_hooks() {
...
+
+	/**
+	 * Loads payment gateway render script if checkout page or pay page.
+	 *
+	 * @since 5.10.5
+	 */
+	public function load_render_js() {
+
+		// also make sure it's not order received page
+		if ( ! is_order_received_page() && ( is_checkout() || is_checkout_pay_page() ) ) {
```
ChaseWiseman, line 114: this is missing a `is_add_payment_method_page()` condition - the JS needs to load on My Account -> Add Payment Method as well.

ChaseWiseman (changes requested):
@N-Molham see my comments regarding some of the JS changes & other considerations.

Also, for a change like this we'll want to make sure and test all of the pages that render the payment form, including:
- Checkout
- Add Payment Method
- Pay page, (by placing a pending order for your user, then going to My Account -> Orders to pay for it)
Checkout is the only page that the bug would be present, but any time the payment form JS is modified in the framework we need to double check all pay pages to make sure nothing else regressed.

Additionally, if you haven't already then I would suggest testing with multiple gateways that have differing form JS:
- Authorize.Net
- CyberSource
We don't need to test all of them, that can be done individually when the framework is updated. But just a couple now to spot check that there are not compatibility considerations.

woocommerce/payment-gateway/assets/js/frontend/sv-wc-payment-gateway-payment-form.coffee
N-Molham: Without this change, the form loads but it doesn't process fields upon submit ... it doesn't load payment fields values, ex => https://cloud.skyver.ge/nOulyjbo

So I thought what other checkout event I can listen to to make sure fields get initialized correctly, so I thought using the `payment_method_selected` is a good idea, so with that change the fields get processed normally.

That said, now when I am testing it, I got a different behavior than yesterday! which I swear it was working!

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
N-Molham: Because `render_js` method gets overridden in multiple gateways, so I thought of a more generic way of handling rendering the JS without the need for the gateways to update their code.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
N-Molham: Yeap, make sense => 2341a97

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
ChaseWiseman: Yeah that's a good point - it does make it easier for gateways to upgrade.

But the main issue still remains and I think is the source of the problem here. `load_render_js` is hooked to `wp_footer`, so when that method fires it will already be after `wp_footer` has been fired. But then it's trying to hook into `wp_footer` again to call `render_js`, but that won't fire again.

It seems to me like this could be simplified and fixed by just calling `render_js()` directly inside `load_render_js()` (though I'd recommend calling it `maybe_render_js()` instead, to keep with what we've done elsewhere). No need for the double `add_action( 'wp_footer'`.

N-Molham: @ChaseWiseman after more digging and testing based on our conversation today, looks like each gateway plugin handles the issue differently, ex: after loading the `render_js` in footer, Bambora throws a ton of console errors but still works! I think these errors can be handled without a problem, but the issue here I believe it comes to each gateway implementation, I don't think there is a solid solution for the FW that fits the majority of the gateways.


https://user-images.githubusercontent.com/3770036/111697121-cede7480-883d-11eb-98c2-24abc340dae8.mp4

tzuk-godaddy:
Just taking a look at this PR here to see if it would address [this CH story](https://app.clubhouse.io/skyverge/story/71253). It doesn't appear like the fix so far matches what I had proposed in the story and shared with the customer back in December as we are still using `wc_enqueue_js()`. However, I suspect that this fix would still work. Could we add testing this with the Conditional Payments plugin to the QA list to ensure that we fix both issues with one fix? Thanks :)

N-Molham:
With this update 4b47a3a (thanks to @tzuk-godaddy guidance) is working ... I tried it with a couple of payment gateways and it seems working ... I know it's not the prettiest solution due to manually adding the `script` tag but I couldn't thing of a better one, let me know what you think

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -99,7 +102,23 @@ protected function add_hooks() {
...
+		add_action( 'wp_footer', [ $this, 'load_render_js' ] );
+	}
+
+
+	/**
+	 * Loads payment gateway render script if checkout page or pay page.
+	 *
+	 * @since 5.10.5
+	 */
+	public function load_render_js() {
```
unfulvio, line 115: I'd rename this method with `maybe_render_js` since it's not loading anything

you could also hook both actions to the same callback and provide a conditional in case the current action is `wp_footer`

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -1013,7 +1032,27 @@ public function render_fieldset_end() {
 	 */
 	public function render_js() {
 
-		wc_enqueue_js( $this->get_safe_handler_js() );
+		// skip if script already rendered
+		if ( true === $this->js_rendered ) {
+			return;
+		}
+
+		$safe_handler_js = '<script type="text/javascript">jQuery(function($) { ' . $this->get_safe_handler_js() . ' });</script>';
```
unfulvio, line 1040: `wc_enqueue_js()` would push this to be loaded in footer anyway, because `wc_print_js()` hooks to `wp_footer`, so I am not sure if I understand the benefit of doing the same thing here - furthermore `wc_print_js` hooks a little later at priority `25`, I am not sure why but could it be so it ensures `jQuery` is enqueued in case it's added in the footer?

what is exactly the advantage of having this hook callback and what problem is trying to solve? the mechanism is not clear to me

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -1013,7 +1032,27 @@ public function render_fieldset_end() {
 	 */
 	public function render_js() {
 
-		wc_enqueue_js( $this->get_safe_handler_js() );
+		// skip if script already rendered
+		if ( true === $this->js_rendered ) {
+			return;
+		}
```
unfulvio, line 1038: I wonder if this simple memoization could hinder problems in the rare occurrence there are multiple gateways initializing different scripts - I know it's unlikely, but maybe the cache key should account for the gateway ID e.g. with `$this->get_gateway()->get_id()`

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -1013,7 +1032,27 @@ public function render_fieldset_end() {
...
+		 * Payment Gateway Payment Form Safe Handler Javascript.
+		 *
+		 * Filters the js code to handler and initialize the payment form.
+		 *
+		 * @since 5.10.7
+		 *
+		 * @param string $safe_handler_js
+		 * @param SV_WC_Payment_Gateway_Payment_Form $this payment form instance
+		 */
+		echo apply_filters( 'wc_' . $this->get_gateway()->get_id() . '_payment_form_safe_handler_js', $safe_handler_js, $this );
```
unfulvio, line 1055: why this method returns a filter? again I am not clear of the purpose

unfulvio (commented):
I'm not 100% sure of the intention of the proposed solution. So if I understood correctly, in the case of a 0$ checkout `"wc_{$gateway_id}_payment_form_end"` is not triggered and hence the JS is not enqueued, is that correct? Is that the reason you're hooking to `wp_footer` instead of `"wc_{$gateway_id}_payment_form_end"` to render JS via `wc_enqueue_js`? If that's so, maybe it'd be helpful to document this. That might explain why you are not using `wc_enqueue_js` any longer given `wc_print_js` also hooks to `wp_footer` so you're in the same context. Would it work hooking  to something just before `wp_footer` so you'd keep consistency with the existing code and method `render_js` using `wc_enqueue_js`? I have left in the PR a few suggestions or comments about the approach, especially WRT memoizing whether the script has rendered or not.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
unfulvio: maybe check my PR review comment, I think there could be a case to still use `wc_enqueue_js` somehow

unfulvio: considering that some plugins may override `render_js` as you pointed elsewhere, how about calling `$this->render_js()` after those checks? if the problem is the hook context, how about using something a bit higher than `wp_footer` but which is ensured to be present in the relevant pages we are interested in?

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -99,7 +102,23 @@ protected function add_hooks() {
...
-		add_action( "wc_{$gateway_id}_payment_form_end",   array( $this, 'render_js' ), 5 );
+		add_action( "wc_{$gateway_id}_payment_form_end", [ $this, 'render_js' ], 5 );
+		add_action( 'wp_footer', [ $this, 'load_render_js' ] );
+	}
+
+
+	/**
+	 * Loads payment gateway render script if checkout page or pay page.
+	 *
+	 * @since 5.10.5
```
unfulvio, line 113:
```suggestion
	 * @since 5.10.7
```

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
nmolham-godaddy: Do you think it's a good idea to change the method name? we have to consider that it will result updating the implementation in all gateway plugins for the new name ... how about after renaming the method, check if the old method exists and call it to introduce a layer of compatibility, ex:
```php
if (method_exists($this, 'load_render_js')) {
	$this->load_render_js();
}
```

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
nmolham-godaddy: The problem with using `wc_enqueue_js` doesn't work due to the same reason you mentioned, the `render_js` gets called when the `wp_footer` action is triggered, it it's matter of race condition.

I tried to offset this by adjusting the `priority` argument to ensure the callback gets called before the `wc_print_js` but with no luck, `add_action( 'wp_footer', [ $this, 'load_render_js' ], 0 );` I am not sure why it doesn't work

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
nmolham-godaddy: do you mean something like a static array property holds a list of gateway IDs of script rendered? ex:
```php
static $js_rendered = [];

public function render_js() {
    if ( in_array( $this->get_gateway()->get_id(), static::$js_rendered, true ) ) {
        returnl
    }
}
```

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
nmolham-godaddy: Yeah it's a new filter I introduced as before the JS didn't have a way for 3rd party to integrate with

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
unfulvio: did `load_render_js` existed in the gateways? I didn't think so

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
unfulvio: something like that, don't think needs to be static though

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
unfulvio: let's not introduce new things if there's no reason for, or it is asked in the implementation

if you introduce something in the public API it means we have to maintain it - unless the filter is part of the fix, that our plugins need to implement I wouldn't suggest to add it in this case

unfulvio (approved):
going to approve this following discussion in 

https://godaddy.slack.com/archives/C01NH3P7FGB/p1621912090041400

will make some changes accordingly to discussion

unfulvio:
I  have added 2 PR for Authorize.Net and CyberSource if those are QA'd successfully I'll merge and tag a new FW version here

Issue #541, opened 2021-05-18
SV_WP_Background_Job_Handler::maybe_handle() throws deprecated notices from Woo 5.3


claudiosanches:
Hello team!

Recently we [fixed an old bug in Woo core](https://github.com/woocommerce/woocommerce/pull/29542), and seems like that this framework is also trying to deal with it, and throws some PHP notices since this framework is calling our old and deprecated `WC()->session->nonce_user_logged_out()` method, now in 5.3 doesn't seems necessary any workaround like this: 
https://github.com/skyverge/wc-plugin-framework/blob/403a2d0d5ad4cf2f1b38675f0f9d41b33353f622/woocommerce/utilities/class-sv-wp-background-job-handler.php#L153-L166

We found out it by some support requests coming from the Facebook integration: https://wordpress.org/support/topic/nonce_user_logged_out-function-is-deprecated/

unfulvio:
hey @claudiosanches thanks for the heads up sorry if this fell under our radar - thankfully another user in #545 brought it up as well and we can get this patched

unfulvio:
issue is fixed in v5.10.13 of the FW, individual plugins that use background jobs will need to update to this version, thanks again for reporting this!

PR #545, opened 2021-08-04
update deprecated function nonce_user_logged_out

glagonikas:
Resolves https://github.com/skyverge/wc-plugin-framework/issues/541 and https://github.com/woocommerce/facebook-for-woocommerce/issues/1966 Other Skyverge plugins like Social Login and Address Validation use this code and should benefit from this change


glagonikas:
@unfulvio @adrianostanley apologies for mentioning you directly. Any change you can review and merge this, so the change propagates to all other Skyverge plugins soon?

unfulvio:
@glagonikas thanks for your PR and the heads up!
It looks like this is solvable with your proposed PR - we just need a backwards compatibility check for older WC versions so either callback can be supported.
We need to update all plugins that utilize this in order to get rid of the deprecated notice.

glagonikas:
@unfulvio PR updated with backwards compatibility

glagonikas:
Ah, I notice you've already made a PR for it. I'll close mine then

unfulvio:
hey sorry about that, yeah I wanted to go with our internal processes, plus utilize some internal helpers - I also noticed that the number of arguments passed from action to callback changed so it needs to account for that

PR #547, opened 2021-08-05
Use updated function in WC_Session_Handler in BG job handler

unfulvio:
## Summary

Updates the BG Job Handler so that it doesn't utilize a deprecated callback function from WC.

## QA

- [ ] Code review

Need to provide concrete replication steps in plugin to make sure change in WC 5.3 didn't affect BG Job Handler mechanism.

The workaround the FW had _may_ not be needed anymore, but that probably requires user testing of several scenarios. By looking at the WC code keeping the same behavior using the updated callback shouldn't affect plugins, we could make a note in the FW comments though.

PR may be tested by bumping framework  in test plugin composer with `dev-mwc-1981/fix-wc-deprecated-notice-in-bg-job-handler`.

### DIFF
--- a/woocommerce/utilities/class-sv-wp-background-job-handler.php
+++ b/woocommerce/utilities/class-sv-wp-background-job-handler.php
@@ -150,20 +150,26 @@ abstract class SV_WP_Background_Job_Handler extends SV_WP_Async_Request {
 			wp_die();
 		}
 
-		// WC core does 2 things here that can interfere with our nonce check:
-		// 1. WooCommerce starts a session due to our GET request to dispatch a job
-		//  However, this happens *after* we've generated a nonce without a session (in CRON context)
-		// 2. it then filters nonces for logged-out users indiscriminately without checking the nonce action; if
-		//  there is a session created (and now the server does have one), it tries to filter every.single.nonce
-		//  for logged-out users to use the customer session ID instead of 0 for user ID. We *want* to check
-		//  against a UID of 0 (since that's how the nonce was created), so we temporarily pause the
-		//  logged-out nonce hijacking before standing aside.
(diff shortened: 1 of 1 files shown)


woocommerce/utilities/class-sv-wp-background-job-handler.php
```diff
@@ -150,20 +150,35 @@ public function maybe_handle() {
...
-		//  against a UID of 0 (since that's how the nonce was created), so we temporarily pause the
-		//  logged-out nonce hijacking before standing aside.
-		remove_filter( 'nonce_user_logged_out', array( WC()->session, 'nonce_user_logged_out' ) );
+		/**
+		 * WC core does 2 things here that can interfere with our nonce check:
+		 *
+		 * 1. WooCommerce starts a session due to our GET request to dispatch a job
+		 *  However, this happens *after* we've generated a nonce without a session (in CRON context)
+		 * 2. it then filters nonces for logged-out users indiscriminately without checking the nonce action; if
+		 *  there is a session created (and now the server does havegit  one), it tries to filter every.single.nonce
```
wvega, line 159: Just a `git` string that sneaked in:

```suggestion
		 *  there is a session created (and now the server does have one), it tries to filter every.single.nonce
```

unfulvio-godaddy: ahah I must have typed that from the console -- thanks for the catch

rawdreeg:
Hi,

Can I have any update on this fix? We've had a few reported [issues](https://github.com/woocommerce/facebook-for-woocommerce/issues/1966) related to this, it'd be good to know when this might get released?

jstern-godaddy (approved):

PR #548, opened 2021-08-13
Creates function for retrieving placeholder lists

ssmith1-godaddy:
# Summary 
Creates helper function for retrieving placeholder lists to be used with wpdb::prepare()

⚠️ This [PR](https://github.com/gdcorp-partners/woocommerce-shipping-local-pickup-plus/pull/54) relies on this PR being merged ⚠️ 

## QA 

- [ ] Code Review
- [x] Unit Test Passes

## Before merge 

- [x] This PR makes the appropriate modifications to automated tests or explains why none are needed
- [ ] This PR makes the appropriate modifications to documentation or explains why none are needed
- [ ] This change does not impact usage or expected outputs of covered code

### DIFF
--- a/tests/unit/HelperTest.php
+++ b/tests/unit/HelperTest.php
@@ -1,0 +1,51 @@
+<?php
+
+define( 'ABSPATH', true );
+
+class HelperTest extends \Codeception\Test\Unit {
+
+	/**
+	 * @var \UnitTester
+	 */
+	protected $tester;
+
+	protected function _before() {
+
+		require_once( 'woocommerce/class-sv-wc-helper.php' );
+	}
+
+	protected function _after() {
+
+	}
+
+	/**
+	 * Tests \SkyVerge\WooCommerce\PluginFramework\v5_10_8\SV_WC_Helper::get_placeholder_list()
+	 *
+	 * @param array $array input country code
+	 * @param string $expected expected return value
+	 *
+	 * @dataProvider provider_can_get_placeholder_list
+	 */
+	public function test_can_get_placeholder_list( array $array, $expected, $placeholder ) {
+
+		$result = \SkyVerge\WooCommerce\PluginFramework\v5_10_8\SV_WC_Helper::get_placeholder_list( $array, $placeholder );
+		$this->assertEquals( $expected, $result );
+	}
+
+
+	/**
+	 * Provider for test_can_get_placeholder_list()
+	 *
+	 * @return array
+	 */
+	public function provider_can_get_placeholder_list() {
+
+		return [
+			'Floating point values' => [ [1.1, 2.1], '%f, %f', '%f' ],
+			'Integer values' => [ [1, 2], '%d, %d', '%d' ],
+			'String values' => [ ['yes', 'no'], '%s, %s', '%s' ],
+		];
+	}
+
+
+}
--- a/woocommerce/class-sv-wc-helper.php
+++ b/woocommerce/class-sv-wc-helper.php
@@ -1095,7 +1095,18 @@ class SV_WC_Helper {
 		}
 	}
 
-
+	/**
+	 * Returns a string of placeholders correlating to the quantity of the passed in array.
+	 *
+	 * This function prepares a string list of placeholders to be utilized with $wpdb::prepare().
+	 *
+	 * @since x.y.z
+	 * @param array $array An array of SQL query args.
+	 * @param string $placeholder Optional. The placeholder type.
+	 */
+	public static function get_placeholder_list( array $array, string $placeholder = '%s' ) : string {
+		return implode( ', ', array_fill( 0, count( $array ), $placeholder ) );
+	}
 }
 
 


unfulvio-godaddy (commented):
I could be wrong, but I believe the suggestion @ragulka had in his audit was for a FW addition not to format an array of placeholders, and rather convert an array of items to a comma separated string values for use in `IN` SQL clauses, which is a pattern found in a handful of plugins (with Memberships the most common probably, and it has added an helper function for that).

Illmar's was merely a suggestion to keep things dry a bit. In the case of your helper function, is there application outside Local Pickup Plus? 

Also, it looks like to me that the proposed helper function doesn't assume the possibility the placeholders might be different because the items in the array have different types (for example in https://github.com/gdcorp-partners/woocommerce-shipping-local-pickup-plus/blob/a8541489f3a5185323682aac037f50133aca8a3c/src/class-wc-local-pickup-plus-pickup-location.php#L578-L614) 

Maybe a proposed function could auto-determine the type and introduce an appropriate placeholder, but I'm not sure if deferring that decision to an helper method is a good idea. 

If the helper you want to introduce only serves LPP at the moment, I suggest moving it into LPP so that a FW update won't be necessary to ship a plugin release.

tests/unit/HelperTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * Tests \SkyVerge\WooCommerce\PluginFramework\v5_10_8\SV_WC_Helper::get_placeholder_list()
+	 *
+	 * @param array $array input country code
+	 * @param string $expected expected return value
+	 *
+	 * @dataProvider provider_can_get_placeholder_list
+	 */
+	public function test_can_get_placeholder_list( array $array, $expected, $placeholder ) {
```
itambek-godaddy, line 29: The docblock here does not match the method signature. Also I would suggest typehinting all params, not just the first.

tests/unit/HelperTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+		$result = \SkyVerge\WooCommerce\PluginFramework\v5_10_8\SV_WC_Helper::get_placeholder_list( $array, $placeholder );
+		$this->assertEquals( $expected, $result );
```
itambek-godaddy, line 32: You could omit the variable assignment here.

```suggestion
		$this->assertEquals(
			$expected,
			\SkyVerge\WooCommerce\PluginFramework\v5_10_8\SV_WC_Helper::get_placeholder_list( $array, $placeholder )
		);
```

woocommerce/class-sv-wc-helper.php
```diff
@@ -1095,7 +1095,18 @@ public static function trigger_error( $message, $type = E_USER_NOTICE ) {
...
 
-
+	/**
+	 * Returns a string of placeholders correlating to the quantity of the passed in array.
+	 *
+	 * This function prepares a string list of placeholders to be utilized with $wpdb::prepare().
+	 *
+	 * @since x.y.z
+	 * @param array $array An array of SQL query args.
+	 * @param string $placeholder Optional. The placeholder type.
```
itambek-godaddy, line 1105: Probably worth documenting the default value here.

tests/unit/HelperTest.php
```diff
@@ -0,0 +1,51 @@
...
+	protected $tester;
+
+	protected function _before() {
+
+		require_once( 'woocommerce/class-sv-wc-helper.php' );
+	}
+
+	protected function _after() {
+
+	}
```
itambek-godaddy, line 19: Is this empty method required?

itambek-godaddy (changes requested):
@ssmith1-godaddy looking good overall, but I'd like a few things to be cleaned up first :)

itambek-godaddy:
@unfulvio-godaddy the helper introduced here seems to be [exactly what I had in mind](https://gist.github.com/itambek-godaddy/c83030f66806f4807b0e08b9b1cf9636#escapingpreparing-values-for-use-in-an-in-clause) - see the 2nd suggested approach. I don't remember seeing any use-cases where we have a list of values where the types are mixed. IIRC, we only had either lists of strings or integers, nothing mixed. The purpose of this helper would be to keep the most common use-case DRY.

If there's a need for mixed placeholders, then that would be something much harder to generalize/abstract, so I think just implementing it in the plugin on such an occasion would make a lot of sense.

ssmith1-godaddy:
Awesome! Thanks for the feedback @itambek-godaddy

unfulvio-godaddy:
alright @itambek-godaddy @ssmith1-godaddy sorry for misunderstanding

if a FW update is needed for this though, since it's am minor enhancement, it would be helpful if we could release a new FW version that also has other improvements, such as #547 

also this PR could target #546 which is the next release branch

(this is not a MWC project so for plugins and FW we should still use release branches)

ragulka:
> it would be helpful if we could release a new FW version that also has other improvements

@unfulvio-godaddy I think that makes perfect sense. Also, given that other plugins could benefit from using this placeholder helper, should we perhaps update these as well? PRP, Teams, Memberships, CSV Export & PDF Vouchers come to mind.

itambek-godaddy:
@ssmith1-godaddy me and @unfulvio-godaddy discussed this placeholder approach in Slack. We agreed that a helper is beneficial, but he explained how it's confusing to have to use the helper to create the placeholders and then later use `$wpdb->prepare()` to replace these placeholders with actual escaped values.

We also realized that Memberships has a `Strings_Helper` class that implements something similar: https://github.com/gdcorp-partners/woocommerce-memberships/blob/master/src/Helpers/Strings_Helper.php#L38-L77 - however we realized the `esc_sql_in_strings` method is broken, as it does not wrap strings in quotes.

In the end, we agreed that a helper that produces a string that can safely be inserted to a SQL statement would be preferred. As explained in the audit, my preference was the `$wpdb->prepare()` approach due to automatic typecasting and string wrapping. We realized we could still use that approach, but prepare the statement inside the helper itself. However, escaping an ID list is much simpler, so in the end we'd want to have 2 distinct methods:

```php
class SV_WC_Helper {

    public static function get_escaped_string_list( array $values ) : string {
        global $wpdb;

        return $wpdb->prepare( implode( ', ', array_fill( 0, count( $values ), '%s' ) ), $values );
    }

    public static function get_escaped_id_list( array $ids ) : string {
        
        return implode( ',', array_unique( array_map( 'absint', $ids ) ) );
    }
}

// Example usage:
$status_list = SV_WC_Helper::get_escaped_string_list(['publish', 'draft']);
$sql = "SELECT * FROM {$wpdb->posts} WHERE post_status IN ( { $status_list } )";

$id_list = SV_WC_Helper::get_escaped_id_list( [1,2,'3'] );
$sql = "SELECT * FROM {$wpdb->posts} WHERE ID IN ( { $id_list } )";

// result:
"SELECT FROM wp_posts WHERE post_status IN ( 'publish', 'draft' )"
"SELECT FROM wp_posts WHERE ID IN ( 1,2,3 )"
```

The obvious benefit with the proposed approach above is that we no longer need to use `$wpdb->prepare()` after getting the list, so it's easier to use and will help avoid bugs.

Let me know if you think this is a better approach and if so, could you try and refactor both PRs with the proposed changes in mind?

PR #552, opened 2021-09-16
Add cacheable request trait

ragulka:
## Summary

Adds a new `CacheableRequestTrait` trait, designed to be used with classes implementing the `SV_WC_API_Request` interface.

## Details

The story requires the trait to provide a method for getting the transient (cache) key for the request, as well as allowing request classes using this trait to customize the params that should be used for generating the transient key. However, I realized that a request instance does not have access to the plugin instance, which would be required to generate the transient key (to avoid potential conflicts with other frameworked plugins). As such, I decided that the transient key should be generated in the `SV_WC_API_Base` class instead.

### Code style considerations

I was unsure what our code style, naming or namespacing requirements are for the framework repo. I see various code styles being used and it seems newer classes are closer to the PSR standards.

At the moment, I have used PSR-compatible file and class naming, while keeping method names underscored.

Please let me know if anything in this regard should be changed.

## Open questions

- [x] Should we add support for toggling caching per request? the `set_force_refresh()` method is used to force a fresh request, but [still stores the response in cache](https://github.com/skyverge/wc-plugin-framework/pull/553/files#diff-62aa93097d1246a64bc9bdc81fdfd4447d6ab85514a681d1772260b83cb80874R130-R132). Would it be useful to have a `set_cacheable()` method to be able to turn off both loading from _and_ storing to the cache?

## QA

- [x] Code review
- [x] Unit tests pass

### DIFF
--- a/tests/unit/API/CacheableRequestTraitTest.php
+++ b/tests/unit/API/CacheableRequestTraitTest.php
@@ -2,9 +2,8 @@
 
 namespace API;
 
-use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\CacheableRequestTrait;
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Cacheable_Request_Trait;
 use SkyVerge\WooCommerce\PluginFramework\v5_10_10\SV_WC_API_JSON_Request;
-use SkyVerge\WooCommerce\PluginFramework\v5_10_10\SV_WC_API_Request;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	define( 'ABSPATH', true );
@@ -22,7 +21,7 @@ class CacheableRequestTraitTest extends \Codeception\Test\Unit
 
 		require_once('woocommerce/api/interface-sv-wc-api-request.php');
 		require_once('woocommerce/api/abstract-sv-wc-api-json-request.php');
-		require_once('woocommerce/api/CacheableRequestTrait.php');
+		require_once('woocommerce/api/Cacheable_Request_Trait.php');
 	}
 
 	/** @see CacheableRequest::get_cache_lifetime() */
@@ -62,7 +61,7 @@ class CacheableRequestTraitTest extends \Codeception\Test\Unit
 	 */
 	protected function get_test_request_instance() : SV_WC_API_JSON_Request {
 		return new class extends SV_WC_API_JSON_Request {
-			use CacheableRequestTrait;
+			use Cacheable_Request_Trait;
 		};
 	}
 }
--- a/tests/unit/API/CacheableRequestTraitTest.php
+++ b/tests/unit/API/CacheableRequestTraitTest.php
@@ -29,7 +29,7 @@ class CacheableRequestTraitTest extends \Codeception\Test\Unit
 	public function test_get_cache_lifetime() {
 		$request = $this->get_test_request_instance();
 
-		$this->assertSame( 0, $request->get_cache_lifetime() );
+		$this->assertSame( 86400, $request->get_cache_lifetime() );
 	}
 
 	/** @see CacheableRequest::set_cache_lifetime() */
--- a/tests/unit/API/CacheableRequestTraitTest.php
+++ b/tests/unit/API/CacheableRequestTraitTest.php
@@ -3,7 +3,6 @@
 namespace API;
 
 use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Cacheable_Request_Trait;
-use SkyVerge\WooCommerce\PluginFramework\v5_10_10\SV_WC_API_Request;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	define( 'ABSPATH', true );
--- a/tests/unit/API/CacheableRequestTraitTest.php
+++ b/tests/unit/API/CacheableRequestTraitTest.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace API;
+
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\CacheableRequestTrait;
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\SV_WC_API_JSON_Request;
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\SV_WC_API_Request;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	define( 'ABSPATH', true );
+}
+
+class CacheableRequestTraitTest extends \Codeception\Test\Unit
+{
+
+
+	/**
+	 * Runs before each test.
+	 */
+	protected function _before()
+	{
+
+		require_once('woocommerce/api/interface-sv-wc-api-request.php');
+		require_once('woocommerce/api/abstract-sv-wc-api-json-request.php');
+		require_once('woocommerce/api/CacheableRequestTrait.php');
+	}
+
+	/** @see CacheableRequest::get_cache_lifetime() */
+	public function test_get_cache_lifetime() {
+		$request = $this->get_test_request_instance();
+
+		$this->assertSame( 0, $request->get_cache_lifetime() );
+	}
+
+	/** @see CacheableRequest::set_cache_lifetime() */
+	public function test_set_cache_lifetime() {
+		$request = $this->get_test_request_instance();
+
+		$this->assertSame( $request, $request->set_cache_lifetime( 1000 ) );
+		$this->assertSame( 1000, $request->get_cache_lifetime() );
+	}
+
+	/** @see CacheableRequest::should_refresh() */
+	public function test_should_refresh() {
+		$request = $this->get_test_request_instance();
+
+		$this->assertFalse( $request->should_refresh() );
+	}
+
+	/** @see CacheableRequest::set_force_refresh() */
+	public function test_set_force_refresh() {
+		$request = $this->get_test_request_instance();
+
+		$this->assertSame( $request, $request->set_force_refresh( true ) );
+		$this->assertTrue( $request->should_refresh() );
+	}
+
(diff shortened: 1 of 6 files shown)


woocommerce/api/CacheableRequestTrait.php
```diff
@@ -0,0 +1,93 @@
...
+ * @copyright Copyright (c) 2013-2021, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10\API;
+
+trait CacheableRequestTrait {
+
+	/** @var int the cache lifetime for the request, in seconds, defaults to 0 (unlimited) */
+	protected $cache_lifetime = 0;
```
ragulka, line 30: I just realized that defining the property in the trait makes it impossible for classes using this property to use property assignment. If a class wants to override the default `$cache_lifetime`, it would need to do so in the constructor, or on a per-instance basis by using `$request->set_cache_lifetime()`.

I don't really fancy the idea of having to defining the default cache lifetime of a request class in it's constructor. However, I'm not excited about not enforcing the property to be protected by the trait, either. Afterall, if the trait does not define the property as protected, then the class using this trait would have free reign over whether it defines the property as protected, public, private or whether it defines the property at all.

As such, I'm torn.

ragulka: Also, I'm not sure if an unlimited cache lifetime by default is reasonable. This might fill up the `wp_options` table rather quickly on busy sites.

ragulka: I went back and forth with these 2 things. In the end, I think it makes sense to provide a sane default that won't blow up anybody's database. With regards to the `$cache_lifetime` property - I think having to assign the property in the constructor is the lesser evil considering the alternative, which is to leave it undefined and up to the class using this trait to define it. The issue with that is that there's nothing enforcing the property to be defined.

woocommerce/api/CacheableRequestTrait.php
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @package   SkyVerge/WooCommerce/API/Response
+ * @author    SkyVerge
+ * @copyright Copyright (c) 2013-2021, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10\API;
+
+trait CacheableRequestTrait {
```
ChaseWiseman, line 27: Let's add a small description here to note what the trait should be used for

ChaseWiseman: Also, regarding code standards: for new classes we've been still using underscore formatting, just without the usual WP file naming standard.

So I think this would be:
- Class name: `Cacheable_Request_Trait`
- Filename: `Cacheable_Request_Trait.php`

woocommerce/api/CacheableRequestTrait.php
ChaseWiseman: >I just realized that defining the property in the trait makes it impossible for classes using this property to use property assignment.

Yeah I often forget about that! I concur that the best experience we can have is to set it in the constructor, and that's what we've been doing in the other platform code as well. I wish we could redefine the value outright at the top of the class, but the constructor is an alright alternative and better than having things like `property_exists` checks, etc...

ChaseWiseman (changes requested):
@ragulka looking good!

>I was unsure what our code style, naming or namespacing requirements are for the framework repo. I see various code styles being used and it seems newer classes are closer to the PSR standards.

I left a comment about the class/filename, but otherwise you are right about the standards and keeping the snake-case methods names and other spacing standards here.

>Would it be useful to have a set_cacheable() method to be able to turn off both loading from and storing to the cache?

Yes I think it would be good to be able to completely bypass both using the stored cache, but also storing the response and hitting the db. I would suggest something like `set_should_cache()` and `should_cache()`. Those could be used in the API base before storage, etc...

ragulka:
@ChaseWiseman thanks - I've addressed the requested changes. I described the `should_cache()` method as determining whether the response should be cached.

This way we have 2 properties for fine-tuning request caching:
* `should_refresh()` - determines whether a fresh request should be made (so the cached response is ignored)
* `should_cache()` - determines whether the response should be stored in cache (_if_ it was a fresh request and the cache lifetime is not expired)

In order to completely ignore caching for a request, one would then do:
```php
$request = new Request();

$request->set_force_refresh( true )->set_should_cache( false );
```

If that's a little too verbose, I also added a `bypass_cache` method that forces a fresh request and disables storing the response in the cache.

```php
$request = new Request();

$request->bypass_cache();
```

Can you let me know if this is what you had in mind and if it makes sense?

tests/unit/API/CacheableRequestTraitTest.php
```diff
@@ -0,0 +1,91 @@
...
+		$this->assertFalse( $request->should_cache() );
+	}
+
+	/**
+	 * Gets a test request instance using the CacheableRequestTrait.
+	 *
+	 * @return SV_WC_API_JSON_Request
+	 */
+	protected function get_test_request_instance() : SV_WC_API_JSON_Request {
+		return new class extends SV_WC_API_JSON_Request {
```
ChaseWiseman, line 87: Should be able to use `$this->getMockForTrait(Cacheable_Request_Trait::class)` instead for this, which would be good to isolate from `SV_WC_API_JSON_Request` and only have the trait loaded since that's the only thing we're testing.

woocommerce/api/Cacheable_Request_Trait.php
```diff
@@ -0,0 +1,153 @@
...
+	protected $should_cache = true;
+
+
+	/**
+	 * Sets the cache lifetime for this request.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @param int $lifetime cache lifetime, in seconds. Set to 0 for unlimited
+	 * @return Cacheable_Request_Trait $this
```
ChaseWiseman, line 55: I believe we can simplify these `$this` return docs to just `self`. (once we can require PHP 7.1, then we'll also declare the return type as `: self {`)

```suggestion
	 * @return self
```

woocommerce/api/Cacheable_Request_Trait.php
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * This trait can be used to add response caching support to API requests.
+ *
+ * It is intended to be used by a class implementing the SV_WC_API_Request interface. Caching itself is handled
+ * by the SV_WC_API_Base class in the perform_request method.
+ *
+ * Simply adding `use Cacheable_Request_Trait;` to a request class will enable caching, but it's also possible to
+ * customize the cache lifetime by setting it in the request constructor.
+ */
+trait Cacheable_Request_Trait {
```
ChaseWiseman, line 36: How about for [consistency with the rest of the FW classes](https://github.com/skyverge/wc-plugin-framework/blob/master/woocommerce/api/class-sv-wc-api-base.php#L29), let's wrap this whole definition in `! trait_exists()`

woocommerce/api/Cacheable_Request_Trait.php
```diff
@@ -0,0 +1,153 @@
...
+ * versions in the future. If you wish to customize the plugin for your
+ * needs please refer to http://www.skyverge.com
+ *
+ * @package   SkyVerge/WooCommerce/API/Response
+ * @author    SkyVerge
+ * @copyright Copyright (c) 2013-2021, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10\API;
```
ChaseWiseman, line 25: Should have caught in my last review - can we put this class in a `Traits` directory? Full namespace being `SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Traits`

woocommerce/api/Cacheable_Request_Trait.php
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Sets whether a fresh request should be attempted, regardless if a cached response is available.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @param bool $value whether to force a fresh request, or not
+	 * @return Cacheable_Request_Trait $this
+	 */
+	public function set_force_refresh( bool $value) {
```
ChaseWiseman, line 86:
```suggestion
	public function set_force_refresh( bool $value ) {
```

ChaseWiseman (changes requested):
@ragulka looks good! A few code standards items, then this can be merged.

>If that's a little too verbose, I also added a bypass_cache method that forces a fresh request and disables storing the response in the cache.

Makes sense!

woocommerce/api/Cacheable_Request_Trait.php
ChaseWiseman: Ah, and looks like we need the standard `defined( 'ABSPATH' ) or exit;` line before that as well

tests/unit/API/CacheableRequestTraitTest.php
ragulka: good idea, thanks!

woocommerce/api/Cacheable_Request_Trait.php
ragulka: sounds good

woocommerce/api/Cacheable_Request_Trait.php
ragulka: Nice catch, thanks!

ChaseWiseman (approved):

PR #553, opened 2021-09-16
Support cacheable requests in API base class

ragulka:
## Summary

This PR adds request/response caching support to the API base class in a transparent manner. This allows caching requests on a per-request basis. Meaning, if a request class declares support for caching, it's also possible to disable request caching for a concrete request by setting using the `$request->set_force_refresh( true )` setter, before performing the request.

## Details

Originally, I had anticipated that the `CacheableRequestTrait` would be responsible for generating the transient (cache) key for the request. However, as the request instance does not have access to the plugin instance, I decided that the transient key should be generated in the `SV_WC_API_Base` class instead.

On a related note - I also did not add a way to filter or override the params which are used to generate the transient key at the moment. I could not think of an immediate use-case for this. Also, it allowed to keep the implementation much simpler for now, since we're just building the key base don the request URI and body. However, I'm open to suggestions.

The overall caching implementation is such that it should not affect any existing plugins in any way. Caching is completely opt-in. 

The raw response is stored in the cache, as this allows normal response handling and broadcasting. However, the raw response will only be stored after it's been successfully parsed. This will prevent storing the response if the request failed or if there was a parsing error. Consequently, it will also protect from corrupted transients, as the response handler will fail with an error, if the cached response data is broken.

Note that a `response_loaded_from_cache` flag is stored on the API instance, which should help debugging.

## Example Usage

```php

class ExchangeRatesRequest extends SV_WC_API_JSON_Request {

    use CacheableRequestTrait;

    public function __construct() {
        $this->cache_lifetime = 600; // override the deafult cache lifetime
    }

    // ...

}

class ExchangeRateAPI extends SV_WC_API_Base {

    public function get_currencies( $refresh = false )
    {
        $request = new ExchangeRatesRequest();

        if ( $refresh ) {
            $request->set_force_refresh( true );
        }

        $this->perform_request( $request );
    }

    // ..
}

$api = new ExchangeRateAPI();

$api->get_currencies(); // this should be a fresh request, the response will be cached
$api->get_currencies(); // this should use the cached response
$api->get_currencies( true ); // this should be a fresh request again, the response will be cached

// override cache timeout on runtime:
$request->set_cache_lifetime( 60 ); // note - this won't invalidate a previous cache with a longer lifetime - should it?

```

### Code style considerations

See my comments in #552

## Open questions

- [x] Should changing the cache lifetime invalidate any previously cached responses? Ie, should the lifetime be part of the cache key MD5 hash?

## QA

- [x] Review and merge #552 first
- [x] Change the base branch to `release/5.10.10`
- [x] Code review
- [x] Unit & integration tests pass `vendor/bin/codecept run unit && vendor/bin/codecept run integration`

Note: it probably makes sense to review this in tandem with https://github.com/gdcorp-partners/woocommerce-avatax/pull/74 as it helps to validate the functionality and ensure code is working as expected (we don't have tests for the API base class 😬 )

### DIFF
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -219,7 +219,7 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	 */
 	protected function save_response_to_cache( array $response ) {
 
-		set_transient( $this->get_request_transient_key(), $response, $this->get_request_cache_lifetime());
+		set_transient( $this->get_request_transient_key(), $response, $this->get_request_cache_lifetime() );
 	}
 
 
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -22,13 +22,14 @@
  * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
  */
 
-namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10;
+namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10\API;
 
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\SV_WC_API_Base;
 use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Traits\Cacheable_Request_Trait;
 
 defined( 'ABSPATH' ) or exit;
 
-if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_10_10\\Abstract_Cacheable_API_Base' ) ) :
+if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_10_10\\API\\Abstract_Cacheable_API_Base' ) ) :
 
 
 /**
@@ -134,7 +135,7 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	 *
 	 * @return bool
 	 */
-	protected function is_request_cacheable() : bool {
+	public function is_request_cacheable() : bool {
 
 		if ( ! in_array( Cacheable_Request_Trait::class, class_uses( $this->get_request() ), true ) ) {
 			return false;
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -192,6 +192,28 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	 *
 	 * @return array
 	 */
+	protected function get_request_data_for_broadcast() : array {
+
+		$request_data = parent::get_request_data_for_broadcast();
+
+		if ( $this->is_request_cacheable() ) {
+			$request_data['force_refresh'] = $this->get_request()->should_refresh();
+			$request_data['should_cache']  = $this->get_request()->should_cache();
+		}
+
+		return $request_data;
+	}
+
+
+	/**
+	 * Gets the response data for broadcasting the request.
+	 *
+	 * Adds a flag to the response data indicating whether the response was loaded from cache.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @return array
+	 */
 	protected function get_response_data_for_broadcast() : array {
 
 		$response_data = parent::get_response_data_for_broadcast();
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -135,7 +135,7 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	 *
 	 * @return bool
 	 */
-	public function is_request_cacheable() : bool {
+	protected function is_request_cacheable() : bool {
 
 		if ( ! in_array( Cacheable_Request_Trait::class, class_uses( $this->get_request() ), true ) ) {
 			return false;
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -1,9 +1,44 @@
 <?php
+/**
+ * WooCommerce Plugin Framework
+ *
+ * This source file is subject to the GNU General Public License v3.0
+ * that is bundled with this package in the file license.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://www.gnu.org/licenses/gpl-3.0.html
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@skyverge.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade the plugin to newer
+ * versions in the future. If you wish to customize the plugin for your
+ * needs please refer to http://www.skyverge.com
+ *
+ * @package   SkyVerge/WooCommerce/API
+ * @author    SkyVerge
+ * @copyright Copyright (c) 2013-2021, SkyVerge, Inc.
+ * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
+ */
 
 namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10;
 
-use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Cacheable_Request_Trait;
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Traits\Cacheable_Request_Trait;
 
+defined( 'ABSPATH' ) or exit;
+
+if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_10_10\\Abstract_Cacheable_API_Base' ) ) :
+
+
+/**
+ * Abstract API base class with caching support.
+ *
+ * Plugins which need to use API request caching should use extend this API base class rather than SV_WC_API_Base.
+ * In addition, each request class which needs caching, should use the Cacheable_Request_Trait.
+ *
+ * @since 5.10.10
+ */
 abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 {
 
@@ -39,7 +74,7 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	/**
 	 * Handle and parse the response
 	 *
-	 * @since 2.2.0
+	 * @since 5.10.10
 	 *
 	 * @param array|\WP_Error $response response data
 	 * @throws SV_WC_API_Exception network issues, timeouts, API errors, etc
@@ -62,7 +97,7 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	/**
 	 * Resets the API response members to their default values.
 	 *
-	 * @since 1.0.0
+	 * @since 5.10.10
 	 */
 	protected function reset_response() {
 
@@ -230,3 +265,6 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	}
 
 }
+
+
+endif;
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -183,4 +183,24 @@ abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
 	}
 
 
+	/**
+	 * Gets the response data for broadcasting the request.
+	 *
+	 * Adds a flag to the response data indicating whether the response was loaded from cache.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @return array
+	 */
+	protected function get_response_data_for_broadcast() : array {
+
+		$response_data = parent::get_response_data_for_broadcast();
+
+		if ( $this->is_request_cacheable() ) {
+			$response_data['from_cache'] = $this->is_response_loaded_from_cache();
+		}
+
+		return $response_data;
+	}
+
 }
--- a/woocommerce/api/Abstract_Cacheable_API_Base.php
+++ b/woocommerce/api/Abstract_Cacheable_API_Base.php
@@ -1,0 +1,186 @@
+<?php
+
+namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10;
+
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\CacheableRequestTrait;
+
+abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
+{
+
+
+	/** @var bool whether the response was loaded from cache */
+	protected $response_loaded_from_cache = false;
+
+
+	/**
+	 * Simple wrapper for wp_remote_request() so child classes can override this
+	 * and provide their own transport mechanism if needed, e.g. a custom
+	 * cURL implementation
+	 *
+	 * @since 2.2.0
+	 *
+	 * @param string $request_uri
+	 * @param string $request_args
+	 * @return array|\WP_Error
+	 */
+	protected function do_remote_request( $request_uri, $request_args ) {
+
+		if ( $this->is_request_cacheable() && ! $this->get_request()->should_refresh() && $response = $this->load_response_from_cache() ) {
+
+			$this->response_loaded_from_cache = true;
+
+			return $response;
+		}
+
+		return parent::do_remote_request( $request_uri, $request_args );
+	}
+
+
+	/**
+	 * Handle and parse the response
+	 *
+	 * @since 2.2.0
+	 * @param array|\WP_Error $response response data
+	 * @throws SV_WC_API_Exception network issues, timeouts, API errors, etc
+	 * @return SV_WC_API_Request|object request class instance that implements SV_WC_API_Request
+	 */
+	protected function handle_response( $response ) {
+
+		parent::handle_response( $response );
+
+		// cache the response
+		if ( ! $this->is_response_loaded_from_cache() && $this->is_request_cacheable() ) {
+
+			$this->save_response_to_cache( $response );
+		}
+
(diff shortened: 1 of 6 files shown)


woocommerce/api/class-sv-wc-api-base.php
```diff
@@ -263,10 +280,11 @@ protected function broadcast_request() {
...
 		$response_data = array(
-			'code'    => $this->get_response_code(),
-			'message' => $this->get_response_message(),
-			'headers' => $this->get_response_headers(),
-			'body'    => $this->get_sanitized_response_body() ? $this->get_sanitized_response_body() : $this->get_raw_response_body(),
+			'code'       => $this->get_response_code(),
+			'message'    => $this->get_response_message(),
+			'headers'    => $this->get_response_headers(),
+			'body'       => $this->get_sanitized_response_body() ? $this->get_sanitized_response_body() : $this->get_raw_response_body(),
+			'from_cache' => $this->is_response_loaded_from_cache(),
```
ragulka, line 287: @ChaseWiseman this produces log messages like the ones below:

When response was not loaded from cache:
```
body: { ... }
from_cache:
```

When response _was_ loaded from cache:
```
body: { ... }
from_cache: 1
```

I'm not particularly fond of the `1` and empty string representation here. It would be more readable if the logs would say `yes|no`, but that is harder to interpret programmatically. Would it make sense to add a formatter around this value [here](https://github.com/skyverge/wc-plugin-framework/blob/146ebc07aeaf8dc11378e51df1b0447cff83fefd/woocommerce/class-sv-wc-plugin.php#L721)?

Also, would it make sense to only add this key to the array if the request class supports caching?

ragulka: I've addressed the above in b3032d4 and e5272cd. Let me know what you think?

woocommerce/api/class-sv-wc-api-base.php
```diff
@@ -548,6 +571,79 @@ protected function get_request_duration() {
...
+
+
+	/**
+	 * Checks whether the current request is cacheable.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @return bool
+	 */
+	protected function is_request_cacheable() : bool {
```
ChaseWiseman, line 601: For methods like these, which are checking or getting a value from the request, I would be more inclined to pass the request object in as an argument rather than always assume we want the request from `get_request()`. That would increase flexibility for APIs that have complex request logic. Granted at the point this is called `get_request()` _should_ return the only request we need to worry about, but again for flexibility if that's not the case in concrete plugins.

That also makes things more testable, should we add tests to for this class as we can test simple input and output and don't need to rely on mocking.

woocommerce/api/class-sv-wc-api-base.php
```diff
@@ -548,6 +571,79 @@ protected function get_request_duration() {
...
+
+
+	/**
+	 * Gets the cache lifetime for the current request.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @return int
+	 */
+	protected function get_request_cache_lifetime() : int {
```
ChaseWiseman, line 631: Same here re: passing in the request object

woocommerce/api/class-sv-wc-api-base.php
```diff
@@ -548,6 +571,79 @@ protected function get_request_duration() {
...
+	 * Request transients can be disabled by using the filter below.
+	 *
+	 * @since 5.10.10
+	 *
+	 * @return string transient key
+	 */
+	protected function get_request_transient_key() : string {
+
+		// ex: wc_sv_requests_<md5 hash of plugin_id, request uri and request data>
+		return sprintf( 'wc_sv_requests_%s', md5( implode( '_', [
```
ChaseWiseman, line 586: any reason these keys need to be generic to the FW? I know the md5 hash will make it unique to the plugin and request technically, but also obfuscates the db entry and makes for harder debugging. What about something like:

`wc_{plugin_id}_api_response_{hash}`? Also noting that the value we're caching here is essentially the response, not the request.

woocommerce/api/class-sv-wc-api-base.php
```diff
@@ -99,21 +103,34 @@ protected function perform_request( $request ) {
...
 
 			// parse & validate response
-			$response = $this->handle_response( $response );
+			$parsed_response = $this->handle_response( $response );
+
+			// cache the response
+			if ( ! $this->is_response_loaded_from_cache() && $this->is_request_cacheable() ) {
+
+				$this->save_response_to_cache( $response );
+			}
```
ChaseWiseman, line 133: Perhaps tuck this at the end of `handle_response()` method instead? Could help cut down on the number of conditionals in this method.

ChaseWiseman (changes requested):
@ragulka looking nice! I left a few comments and suggestions/questions.

>Should changing the cache lifetime invalidate any previously cached responses? Ie, should the lifetime be part of the cache key MD5 hash?

I think that makes sense. That shouldn't happen too often, but no harm in clearing things if it does.

As for the overall approach, I think one potential improvement I see could be made is separation of concerns. Right now the API base class has one responsibility: make API requests and return their result. Adding the caching here gives it two.

In the newer platform code we've done things like have a public repository for a resource e.g.:
`TaxRates::getRates()`

Then inside that we handle the logic of either pulling from cache or making a new API request. The API request handling itself knows nothing about caches or how that's handled.

That said, I can see something like that causing a whole lot of refactoring necessary both here and in a plugin that already handles API requests in a certain way and has to switch to new handlers.

So, as a happy medium, what about a wrapper API base that adds this caching functionality on top of `SV_WC_API_Base`, so there are no changes to the latter and the new wrapper only contains a few overrides for the caching functionality? Then in addition to using the trait, plugins would switch to extending the wrapper.

`Cacheable_API_Base`, for example.

I _think_ that could then override `do_remote_request()` and cache before making the request, then override `handle_response` to store the cache in the same way this PR does. I could be mistaken and there may be some logic that prevents injecting the cache behavior where we need, but at first read-through it seems possible.

What do you think?

woocommerce/api/class-sv-wc-api-base.php
ragulka: I think having the plugin_id there makes perfect sense, thanks! And yeah, naming the key like this makes it clear that it's the response that's cached.

ragulka:
> So, as a happy medium, what about a wrapper API base that adds this caching functionality on top of `SV_WC_API_Base`, so there are no changes to the latter and the new wrapper only contains a few overrides for the caching functionality? Then in addition to using the trait, plugins would switch to extending the wrapper.
> 
> `Cacheable_API_Base`, for example.
> 
> I _think_ that could then override `do_remote_request()` and cache before making the request, then override `handle_response` to store the cache in the same way this PR does. I could be mistaken and there may be some logic that prevents injecting the cache behavior where we need, but at first read-through it seems possible.

@ChaseWiseman I agree on your points above about class size, responsibility, etc and also concur on trying to avoid a larger than necessary refactor here and across our plugins.

I think the proposal to tuck cache handling to `do_remote_request` and `handle_response` could work, yeah. And then, moving this as well as all other cache-related functionality to a separate class, or perhaps a trait?

woocommerce/api/class-sv-wc-api-base.php
ragulka: Interesting, I was following the current pattern in this class, which is that we don't seem to pass around the request to any methods that deal with the request (ie, `get_request_uri`, `get_request_body`, etc).

Do you have any concrete plugins / use-cases in mind where the current implementation would be a problem? Or is it something we're looking to change for these other methods in the future as well?

ragulka:
@ChaseWiseman I've implemented your suggestions. I decided to go with an abstract class instead of a trait because this will make it easier for child classes to override any methods while still being able to call parent methods. When using a trait, one would have to specifically `use as` when overriding a trait method and wanting to call it in the overridden method.

I also extracted the request and response data setup to dedicated methods when broadcasting a request. This allows child classes to customize this data, which is something that I'm doing in the `Abstract_Cacheable_API_Base` to add caching flags to request logging:

```
Request
method: GET
// ... stripped for brevity
force_refresh: no
should_cache: yes

Response
code: 200
// ... stripped for brevity
from_cache: yes
```

woocommerce/api/class-sv-wc-api-base.php
ChaseWiseman: no specific use cases in mind, just a general preference that I've developed as we've worked on more modern projects. You're certainly right that the approach is consistent with existing handling. I doubt we'll go back and change things in this class any time in the near future. No big deal to leave as-is just for consistency.

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
+<?php
+
```
ChaseWiseman, line 2: Needs the plugin framework header

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+namespace SkyVerge\WooCommerce\PluginFramework\v5_10_10;
+
```
ChaseWiseman, line 4: Needs the `defined( 'ABSPATH' ) or exit;` bail

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+use SkyVerge\WooCommerce\PluginFramework\v5_10_10\API\Cacheable_Request_Trait;
+
+abstract class Abstract_Cacheable_API_Base extends SV_WC_API_Base
```
ChaseWiseman, line 7: At a minimum let's make sure and add a `@since` tag for this, and a description if one is useful

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+	/** @var bool whether the response was loaded from cache */
+	protected $response_loaded_from_cache = false;
+
+
+	/**
+	 * Simple wrapper for wp_remote_request() so child classes can override this
+	 * and provide their own transport mechanism if needed, e.g. a custom
+	 * cURL implementation
+	 *
+	 * @since 2.2.0
```
ChaseWiseman, line 20: Should this be set to when the override was added? Since we're in the context of this class.
```suggestion
	 * @since 5.10.10
```

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		return parent::do_remote_request( $request_uri, $request_args );
+	}
+
+
+	/**
+	 * Handle and parse the response
+	 *
+	 * @since 2.2.0
```
ChaseWiseman, line 42:
```suggestion
	 * @since 2.2.0
	 *
```

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+
+			$this->save_response_to_cache( $response );
+		}
+
+		return $this->response; // this param is set by the parent method
+	}
+
+
+	/**
+	 * Reset the API response members to their
```
ChaseWiseman, line 62:
```suggestion
	 * Resets the API response members to their default values.
```

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+		/**
+		 * Filters whether the API request is cacheable.
+		 *
+		 * Allows actors to disable API request caching when a request is normally cacheable. This may be useful
+		 * primarily for debugging situations.
+		 *
+		 * Note: this filter is only applied if the request is originally cacheable, in order to prevent issues when
+		 * a non-cacheable request is accidentally flagged as cacheable.
+		 *
+		 * @since 5.10.10
```
ChaseWiseman, line 116:
```suggestion
		 * @since 5.10.10
		 *
```

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+	 */
+	protected function get_request_cache_lifetime() : int {
+
+		/**
+		 * Filters API request cache lifetime.
+		 *
+		 * Allows actors to override cache lifetime for cacheable API requests. This may be useful for debugging
+		 * API requests by temporarily setting short cache timeouts.
+		 *
+		 * @since 5.10.10
```
ChaseWiseman, line 139:
```suggestion
		 * @since 5.10.10
		 *
```

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+		 */
+		return (int) apply_filters( 'wc_plugin_' . $this->get_plugin()->get_id() . '_api_request_cache_lifetime' , $this->get_request()->get_cache_lifetime(), $this->get_request() );
+	}
+
+
+
+	/**
+	 * Determine whether the response was loaded from cache or not.
+	 *
+	 * @since 5.10.10
```
ChaseWiseman, line 151:
```suggestion
	 * @since 5.10.10
	 *
```

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+	 *
+	 * @return array
+	 */
+	protected function get_request_data_for_broadcast() : array {
+
+		$request_data = parent::get_request_data_for_broadcast();
+
+		if ( $this->is_request_cacheable() ) {
+			$request_data['force_refresh'] = $this->get_request()->should_refresh();
```
ChaseWiseman, line 200: In practice I find these two flags to be useful when a non-cached request is made, but confusing when the cache is used since they indicate that a request is actually made when it wasn't.

What if we conditionally add the `from_cache` line to this as well? So if a real request is indeed being made, the `force_refresh` and `should_cache` flags are added. But, if we're not going to be making a request and the cache response will be used, only the `from_cache` is added.

That might help reduce confusing in debugging when these requests/responses are logged as if API communication is happening when it's really not.

woocommerce/api/Abstract_Cacheable_API_Base.php
```diff
@@ -0,0 +1,228 @@
...
+	 * @since 5.10.10
+	 *
+	 * @return array
+	 */
+	protected function get_response_data_for_broadcast() : array {
+
+		$response_data = parent::get_response_data_for_broadcast();
+
+		if ( $this->is_request_cacheable() ) {
+			$response_data['from_cache'] = $this->is_response_loaded_from_cache();
```
ChaseWiseman, line 222: Additionally, for this and the request flags, can we bump them up to the top of the array so it's more visible among the other properties?

ChaseWiseman (changes requested):
@ragulka very cool, nice work! The abstraction looks good. I've left some comments for code cleanup and logging approach.

I also tested a few of the request cases listed in https://github.com/gdcorp-partners/woocommerce-avatax/pull/74 and the caching seems to be working properly (heads up that that PR needs the composer.lock updated with these latest changes, needed to do that manually to test it. There are also some specifics about the option names that could be updated to avoid confusion).

All that's left to do here aside from the code tweaks is to add tests. I believe `$this->getMockForAbstractClass()` will be helpful for this.

No need to go back and add tests for the whole `SV_WC_API_Base` since that doesn't already have them and adding them is a bit out of scope. However, simple tests could be added for the two new broadcast methods introduced here.

Sound good?

woocommerce/api/Abstract_Cacheable_API_Base.php
ragulka: > confusing when the cache is used since they indicate that a request is actually made when it wasn't.

Interesting, my goal with adding these was to _avoid_ confusion when debugging 😅. The idea was that it would be helpful to see whether these flags were properly respected - ie, we'd like to see when a request was allowed to be loaded or stored in cache in the first place. And then later, whether the response was actually loaded from cache or not.

I'm not sure I quite follow you on the "they indicate that a request is actually made" part. Do you mean the request itself should not be logged at all?

woocommerce/api/Abstract_Cacheable_API_Base.php
ragulka: Sure thing

woocommerce/api/Abstract_Cacheable_API_Base.php
cwiseman-godaddy: >I'm not sure I quite follow you on the "they indicate that a request is actually made" part. Do you mean the request itself should not be logged at all?

I think logging the request is fine, but I meant that a flag noting that it wasn't a real API request is probably helpful. I'm thinking of it more as a support engineer, debugging a site that is having trouble with tax calculation and going to the logs to see what communication took place. Everything about the logged request details indicates that there was communication with the API when there may not have been. Seeing `should_cache` makes me think that a request is being made, then cached.

ragulka: > Everything about the logged request details indicates that there was communication with the API when there may not have been. Seeing `should_cache` makes me think that a request is being made, then cached.

Would you say the `from_cache` flag on the logged response is then insufficient in indicating the request was not actually made, but the response was loaded from cache?

ragulka:
@ChaseWiseman I've added tests for the new API base class. Note that I ended up adding integration tests, as there's quite a bit of integration with WP and the other underlying classes. It can be argued that some of these tests could be extracted to a unit test class, but I kept everything together for now.

Let me know what you think.

ChaseWiseman (approved):
LGTM!

woocommerce/api/Abstract_Cacheable_API_Base.php
ChaseWiseman: imo yes. It tells me something about the response, but not the request. The flags added to the request seem to conflict with that notion in my mind - if the cache was used and the request was never made, I wouldn't expect to see things like "should cache" in the logged request, because there wasn't a request made to cache.

I still approved the PR because I don't feel that strongly about it, and it doesn't need to be a blocker right now. We can tweak those messages if there is any confusion when the plugin PRs are QAed.

ragulka: Ah, I think I see your point now. Would it make sense to log the `should_cache` flag in the response data, then?

PR #559, opened 2022-03-22
Drop deprecated SV WC compatibility classes

unfulvio:
## Summary

This PR drops the following deprecated classes:

* `SV_WC_Compatibility` abstract
* `SV_WC_Datetime`
* `SV_WC_Order_Compatibility`
* `SV_WC_Product_Compatibility`

These classes were introduced during WC 3.0 transition to data stores are no longer used in FW since v5.5.0 (2019) and should not affect plugins.

## QA

- [ ] Code review

## After merge

- Create a new release branch of the FW bumping namespaces and version numbers
- Update version of the FW used in MWC features


wvega (changes requested):
@unfulvio thank you for putting these changes together. I agree that it's time to remove those deprecated methods and classes, fixing those PHP 8.1 notices in the process.

I noticed that `SV_WC_Order_Compatibility::get_item_formatted_meta_data()` is not deprecated and still being used in `SV_WC_Helper`. I think we need to keep that one.

wvega:
Also, I found [some deprecated methods](https://github.com/skyverge/wc-plugin-framework/blob/0c79b5799955c59e644caf1da222c20c982eba06/woocommerce/class-sv-wc-plugin-compatibility.php#L54-L121) that reference one of these classes in the docblock. Since I think we need to make a major release for this changes, I think we should go ahead and remove everything that we deprecated in 5.*.

If that sounds good, I can open a separate PR for that.

What do you think?

unfulvio:
hey @wvega good catch! didn't notice that - have updated and added some migration notes to include in the wiki post-release

I also like your proposal to remove more deprecated stuff; I have dropped everything that was deprecated by v5.5.0, which should be reasonable.

unfulvio-godaddy:
fixes #590

PR #562, opened 2022-04-26
Fix payment method image lookup in plugin assets

### LINKED ISSUE #561: Not possible to override credit card icons (or add new supported credit card type icon)
Looks like these lines of code expected to allow plugin developers add their custom credit card icons, within the plugin (first `if`), then using a fallback icon from the framework (second `if`)

https://github.com/skyverge/wc-plugin-framework/blob/0c79b5799955c59e644caf1da222c20c982eba06/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php#L1755-L1763

But in fact, both ifs are equal and point to the same framework assets path. This makes impossible to add new supported credit card type or change the icons for existing types.

cadic:
This PR fixes the lookup of payment method images as expected:

1. Trying to find the image in `plugin/assets/images/card-<method_name>.ext`
2. If the first image doesn't exist, look in the framework assets folder

Closes #561

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -1753,8 +1753,8 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 		$image_extension = apply_filters( 'wc_payment_gateway_' . $this->get_plugin()->get_id() . '_use_svg', true ) ? '.svg' : '.png';
 
 		// first, is the card image available within the plugin?
-		if ( is_readable( $this->get_plugin()->get_payment_gateway_framework_assets_path() . '/images/card-' . $image_type . $image_extension ) ) {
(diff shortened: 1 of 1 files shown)


unfulvio (approved):

unfulvio:
hey @cadic thanks for submitting this and apologies for the late reply - we haven't been very thorough in monitoring this repository lately, but definitely the framework powers all our plugins and some more. This looks like an ancient bug introduced a whole 8 years ago: https://github.com/godaddy-wordpress/wc-plugin-framework/commit/dce86b4580c1ffdbd01d8561d833776dc539ae38#diff-bb4c623d244ef221636588f979faec9da53b449cfd94d997e6c7308e173a1afb - sounds like we never had the case of a plugin of ours overriding the card images and that's why we probably never saw this issue before -- good find!

I'm going to merge this PR into #574 - I'm also going to take a look to an unrelated PR and then later release this - your plugin will have to use the release version of the framework to make use of your fix

Thanks again

Fulvio

PR #581, opened 2023-01-09
HPOS Compatibility helper methods

sahmed2-godaddy:
# Summary
This PR contains adds helper methods for HPOS-related functionalities.

### Details
- Added a helper method to verify if the HPOS functionality is available or not.
- Added CRUD methods for the modifying orders meta

### DIFF
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -500,7 +500,7 @@ class SV_WC_Plugin_Compatibility {
 	 */
 	public static function is_hpos_enabled() {
 
-			return class_exists( $custom_order_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class ) && get_option( $custom_order_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
+			return class_exists( $custom_order_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class ) && 'yes' === get_option( $custom_order_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
 		}
 }
 
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -492,7 +492,9 @@ class SV_WC_Plugin_Compatibility {
 	}
 
 	/**
-	 * Whether the HPOS is enabled.
+	 * Determines whether HPOS is enabled.
+	 *
+	 * @link https://woocommerce.com/document/high-performance-order-storage/
 	 *
 	 * @since x.y.z
 	 *
@@ -500,7 +502,10 @@ class SV_WC_Plugin_Compatibility {
 	 */
 	public static function is_hpos_enabled() {
 
-		return class_exists( $custom_order_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class ) && 'yes' === get_option( $custom_order_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
+		$custom_orders_table_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class;
+
+		return class_exists( $custom_orders_table_controller )
+			&& 'yes' === get_option( $custom_orders_table_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
 	}
 }
 
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -24,6 +24,8 @@
 
 namespace SkyVerge\WooCommerce\PluginFramework\v5_10_14;
 
+use Automattic\WooCommerce\Utilities\OrderUtil;
+
 defined( 'ABSPATH' ) or exit;
 
 if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_10_14\\SV_WC_Plugin_Compatibility' ) ) :
@@ -491,6 +493,7 @@ class SV_WC_Plugin_Compatibility {
 		return $version === $subscriptions_version || ( $subscriptions_version && version_compare( $version, $subscriptions_version, '=' ) );
 	}
 
+
 	/**
 	 * Determines whether HPOS is enabled.
 	 *
@@ -500,13 +503,12 @@ class SV_WC_Plugin_Compatibility {
 	 *
 	 * @return bool
 	 */
-	public static function is_hpos_enabled() {
+	public static function is_hpos_enabled() : bool {
 
-		$custom_orders_table_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class;
-
-		return class_exists( $custom_orders_table_controller )
-			&& 'yes' === get_option( $custom_orders_table_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
+		return OrderUtil::custom_orders_table_usage_is_enabled();
 	}
+
+
 }
 
 
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * WooCommerce Plugin Framework.
+ * WooCommerce Plugin Framework
  *
  * This source file is subject to the GNU General Public License v3.0
  * that is bundled with this package in the file license.txt.
@@ -16,6 +16,7 @@
  * versions in the future. If you wish to customize the plugin for your
  * needs please refer to http://www.skyverge.com
  *
+ * @package   SkyVerge/WooCommerce/Plugin/Classes
  * @author    SkyVerge
  * @copyright Copyright (c) 2013-2022, SkyVerge, Inc.
  * @license   http://www.gnu.org/licenses/gpl-3.0.html GNU General Public License v3.0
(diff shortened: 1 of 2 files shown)


woocommerce/class-sv-wc-plugin-compatibility.php
```diff
@@ -491,7 +491,17 @@ protected static function is_wc_subscriptions_version( $version ) {
 		return $version === $subscriptions_version || ( $subscriptions_version && version_compare( $version, $subscriptions_version, '=' ) );
 	}
 
+	/**
+	 * Whether the HPOS is enabled.
```
unfulvio-godaddy, line 495:
```suggestion
	 * Determines whether the HPOS is enabled.
```

Prefer descriptive titles indicating what the method does in third person

woocommerce/class-sv-wc-plugin-compatibility.php
```diff
@@ -491,7 +491,17 @@ protected static function is_wc_subscriptions_version( $version ) {
...
+	 *
+	 * @since x.y.z
+	 *
+	 * @return bool
+	 */
+	public static function is_hpos_enabled() {
 
+		return class_exists( $custom_order_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class ) && 'yes' === get_option( $custom_order_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
```
unfulvio-godaddy, line 503:
```suggestion
		$custom_orders_table_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class;
		
		return class_exists( $custom_orders_table_controller )
			&& 'yes' === get_option( $custom_order_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
```

This is a rather long line, maybe can be shortened for readability

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+		} else {
+			$value = get_post_meta( $order_id, $key, $single );
+		}
+
+		return $value;
+	}
+
+	/**
+	 * Updates the order meta according to HPOS availability.
+	 * Uses WC_Order data store object if HPOS is enabled, otherwise it uses the traditional update_post_meta() method.
```
unfulvio-godaddy, line 562: add newline between title and description in phpdoc

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 *
+	 * @since x.y.z
+	 *
+	 * @param int $order_id Required. Order ID.
+	 * @param string $key Required. Meta key.
+     * @param mixed $value Required. Meta value.
+     *
+	 */
```
unfulvio-godaddy, line 570: mixed tabs and spaces
remove empty phpdoc line

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * 			   Default false.
+     *
+	 */
+	public static function add_order_meta( int $order_id, string $key, $value, $unique = false ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			$order = wc_get_order($order_id);
+			if( $order ){
```
unfulvio-godaddy, line 606: see my other comment

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+		$value = false;
+
+		// setting $single to false by default if hpos is disabled to match it with get_post_meta method's default value
+		if( is_null($single) ) {
+			$single = $is_hpos_enabled;
+		}
+
+		if ( $is_hpos_enabled ) {
+			$order = wc_get_order($order_id);
+			if( $order ){
```
unfulvio-godaddy, line 550: see my other comment

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 */
+	public static function get_order_meta( int $order_id, string $key, $single = null) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+		$value = false;
+
+		// setting $single to false by default if hpos is disabled to match it with get_post_meta method's default value
+		if( is_null($single) ) {
+			$single = $is_hpos_enabled;
+		}
```
unfulvio-godaddy, line 546: arguments in functions need spaces in between parenthesis

prefer `null === $var` than using function call `is_null`

unfulvio-godaddy: also, I'm not sure why we are setting `$is_hpos_enabled` to $single, while methods invoking this method can set their preference; this would cause their preference to be overridden

could you explain the reasoning behind this?

it seems to  me we could as well default this to true though, and drop the $single argument since most of the use cases assume true WRT post meta

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	/**
+	 * Deletes the order meta according to HPOS availability.
+	 * Uses WC_Order data store object if HPOS is enabled, otherwise it uses the traditional delete_post_meta() method.
+	 *
+	 * @since x.y.z
+	 *
+	 * @param int $order_id Required. Order ID.
+	 * @param string $key Required. Meta key.
+	 * @param mixed $meta_value Optional. Applicable only if the HPOS is inactive.
+     *
```
unfulvio-godaddy, line 624: same issue here

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 */
+	public static function delete_order_meta( int $order_id, string $key ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			$order = wc_get_order($order_id);
+			if( $order ){
```
unfulvio-godaddy, line 632: see my other comment

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+			}
+		} else {
+			update_post_meta( $order_id, $key, $value );
+		}
+	}
+
+
+	/**
+	 * Adds the order meta according to HPOS availability.
+	 * Uses WC_Order data store object if HPOS is enabled, otherwise it uses the traditional add_post_meta() method.
```
unfulvio-godaddy, line 589: see my other comment

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 */
+	public static function add_order_meta( int $order_id, string $key, $value, $unique = false ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			$order = wc_get_order($order_id);
+			if( $order ){
+				$order->add_meta_data( $key, $value, $unique );
+				$order->save();
```
unfulvio-godaddy, line 608: I don't think we need to save the whole order

save_meta_data() would suffice

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 */
+	public static function delete_order_meta( int $order_id, string $key ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			$order = wc_get_order($order_id);
+			if( $order ){
+				$order->delete_meta_data( $key );
+				$order->save();
```
unfulvio-godaddy, line 634: no need for saving the whole order object

unfulvio-godaddy (changes requested):

woocommerce/compatibility/class-sv-wc-order-compatibility.php
sahmed2-godaddy: @unfulvio-godaddy yes, this makes sense but I am following this HPOS guide [here](https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#apis-for-gettingsetting-posts-and-postmeta), so let me know if that's fine or should I replace it with save_meta_data()?

woocommerce/compatibility/class-sv-wc-order-compatibility.php
sahmed2-godaddy: @unfulvio-godaddy yes, I did not set a default value in the parameter because we'd be using this method for both with/without HPOS functionality, so the default value for `$single` in [get_post_meta](https://developer.wordpress.org/reference/functions/get_post_meta/) is `false`, but in `\WC_Order->get_meta()` the default value of `$single` is set to `true`, so I set these default values accordingly to avoid the ambiguity while replacing `get_post_meta()` with this new method in our plugins, so that we can keep the exact same parameters.

Does that makes sense?

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: hmmm, normally save_meta_data() is sufficient, but I wonder why they suggest that - while they also add

> Calling the save() method is a relatively expensive operation, so you may wish to avoid calling it more times than necessary (for example, if you know it will be called later in the same flow, you may wish to avoid additional earlier calls when operating on the same object).

so I wonder if there's something special under the hood when it comes to HPOS and `save()` is necessary

we should investigate if this assumption is correct, or really there's no difference, and save_meta_data is sufficient

if it's not, I would still not place `save()` here - because in the scenario one is setting multiple meta data at once, then this would become a very expensive operation (say one is handling meta in a foreach loop, then save should be called at the very end) - in that case we need a different strategy to have plugins call save when appropriate

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: follow up discussion: https://github.com/godaddy-wordpress/wc-plugin-framework/pull/581#discussion_r1065926128

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,121 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 */
+	public static function update_order_meta( int $order_id, string $key, $value ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			$order = wc_get_order($order_id);
+			if( $order ){
+				$order->update_meta_data($key, $value, $order_id);
+				$order->save();
```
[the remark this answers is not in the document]
unfulvio-godaddy: follow up discussion: follow up discussion: https://github.com/godaddy-wordpress/wc-plugin-framework/pull/581#discussion_r1065926128

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,117 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+		$value = false;
+
+		// setting $single to false by default if hpos is disabled to match it with get_post_meta method's default value
+		if( null === $single ) {
+			$single = $is_hpos_enabled;
+		}
+
+		if ( $is_hpos_enabled ) {
+			if( $order = wc_get_order( $order_id ) ){
```
unfulvio-godaddy, line 550: needs space after if as `if (`

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,117 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * @param int $order_id Required. Order ID.
+	 * @param string $key Required. Meta key.
+	 * @param mixed $value Required. Meta value.
+	 */
+	public static function update_order_meta( int $order_id, string $key, $value ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			if( $order = wc_get_order( $order_id ) ){
```
unfulvio-godaddy, line 576: needs space between if and parenthesis

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,117 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * @param string $key Required. Meta key.
+	 * @param mixed $value Required. meta value.
+	 * @param bool $unique Optional. Whether the same key should not be added. Defaults to false.
+	 */
+	public static function add_order_meta( int $order_id, string $key, $value, $unique = false ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			if( $order = wc_get_order( $order_id ) ){
```
unfulvio-godaddy, line 603: needs space between if and parenthesis

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,117 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * @param int $order_id Required. Order ID.
+	 * @param string $key Required. Meta key.
+	 * @param mixed $meta_value Optional. Applicable only if the HPOS is inactive.
+	 */
+	public static function delete_order_meta( int $order_id, string $key ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			if( $order = wc_get_order( $order_id ) ){
```
unfulvio-godaddy, line 628: needs space between if and parenthesis

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: Upon suggestion of @jstern-godaddy I have contacted the author of that wiki page directly to ask for clarifications.

In the meantime could you leave this to `save_meta_data` and later experiment in QA/testing to see if there's any drawback going with that?

thank you

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,117 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 *             Defaults to true if HPOS is enabled, otherwise false.
+	 *
+	 * @return mixed
+	 */
+	public static function get_order_meta( int $order_id, string $key, $single = null) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+		$value = false;
+
+		// setting $single to false by default if hpos is disabled to match it with get_post_meta method's default value
```
jstern-godaddy, line 544: Is this comment entirely accurate? It says "setting $single to false by default if hpos is disabled" but won't this also set $single to true if HPOS is _enabled_? Is that intended?

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,117 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			if( $order = wc_get_order( $order_id ) ){
+				$order->delete_meta_data( $key );
+				$order->save();
+			}
+		} else {
+			delete_post_meta( $order_id, $key, $meta_value );
```
unfulvio-godaddy, line 633: delete_post_meta does not require `$meta_value`  - this variable is not present and would cause a fatal error

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: so, it turns out this is due to a WC core bug: https://godaddy.slack.com/archives/C026JPZ4GLB/p1673409266503829

this won't halt further work with HPOS on our side however we would prefer to have that sorted before we deploy any plugins that include HPOS compatibility

I would say that for now we still use `save_meta_data` and we come back to this as soon there's a resolution from WC. We hold any FW release or plugin deploy before we are sure. It's also likely we will need an additional condition check here https://github.com/godaddy-wordpress/wc-plugin-framework/pull/582#discussion_r1066554703 before declaring compatibility

sahmed2-godaddy: Understood, thanks for doing the research and sharing these details. I am on it!

woocommerce/compatibility/class-sv-wc-order-compatibility.php
sahmed2-godaddy: @jstern-godaddy please take a look at this above comment thread [here](https://github.com/godaddy-wordpress/wc-plugin-framework/pull/581#discussion_r1065946813) and let me know if I should just set it to false by default in our new method? I did this to avoid any issues while replacing in the plugins.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
sahmed2-godaddy: @unfulvio-godaddy according to this function [reference here](https://developer.wordpress.org/reference/functions/delete_post_meta/), it accepts the third $meta_value parameter which is by default set to an empty string as I did in my method, but yes I don't think we're using this third parameter anywhere, but I added it to keep it same as core method. 

Let me know if I should remove it?

woocommerce/compatibility/class-sv-wc-order-compatibility.php
sahmed2-godaddy: @unfulvio-godaddy yes makes sense, as we're getting a single value in most of the use-cases so I will just set it to `true` by default in the arguments to avoid any confusions by anyone using this method.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: I almost forgot that function supported a 3rd arg actually. The variable wasn't available in the method's logic when I first reviewed. Anyhow, sounds like a bit like https://github.com/godaddy-wordpress/wc-plugin-framework/pull/581#discussion_r1066512178 -- I doubt we use delete_post_meta in our codebase with ever passing the 3rd arg - sounds good leaving the default value as you did; this will have no effect with HPOS on so I don't know if we need it at all. Fine for me to leave as is.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: sounds good to me, thanks

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +522,111 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * @param string $key Required. Meta key.
+	 * @param mixed $value Required. Meta value.
+	 */
+	public static function update_order_meta( int $order_id, string $key, $value ) {
+
+		$is_hpos_enabled = SV_WC_Plugin_Compatibility::is_hpos_enabled();
+
+		if ( $is_hpos_enabled ) {
+			if ( $order = wc_get_order( $order_id ) ) {
+				$order->update_meta_data( $key, $value, $order_id );
```
unfulvio-godaddy, line 571: I believe we don't need to pass the order_id to `update_meta_data` - this is the meta id not the order ID; we could omit this argument entirely since it's not used with `update_post_meta()`

unfulvio-godaddy (changes requested):

woocommerce/compatibility/class-sv-wc-order-compatibility.php
jstern-godaddy: Yeah I'm not certain that I've _ever_ used the non-single version, it never really made much sense to me

jstern-godaddy: > prefer `null === $var` than using function call `is_null`

@unfulvio-godaddy also FWIW as @maxrice pointed out to me the other day, WC coding standards have moved away from yoda conditionals: https://developer.woocommerce.com/2022/08/11/goodbye-yoda-conditions/

woocommerce/compatibility/class-sv-wc-order-compatibility.php
shehrozsheikh: @unfulvio-godaddy nice catch, fixed it.

unfulvio-godaddy (commented):
One thing I'm thinking after reviewing the compatibility methods for handling order meta, is that maybe we should allow passing a `WC_Order` object to the method, alternatively to an ID - I'm generally not a fun of function arguments allowing multiple types but in this scenario it would allow passing a WC_Order object than having to fetch it again with `wc_get_order()` and save some query performance.

woocommerce/class-sv-wc-plugin-compatibility.php
```diff
@@ -491,7 +491,22 @@ protected static function is_wc_subscriptions_version( $version ) {
...
+	 *
+	 * @return bool
+	 */
+	public static function is_hpos_enabled() {
 
+		$custom_orders_table_controller = \Automattic\WooCommerce\Internal\DataStores\Orders\CustomOrdersTableController::class;
+
+		return class_exists( $custom_orders_table_controller )
+			&& 'yes' === get_option( $custom_orders_table_controller::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION );
+	}
```
unfulvio-godaddy, line 509: why we went with this solution and not the one described in https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#detecting-whether-hpos-tables-are-being-used-in-the-store ?

unfulvio-godaddy (approved):

PR #582, opened 2023-01-10
Default HPOS support added for all gateway in framework

sharazghouri:
## Summary

we can automate the way a plugin utilizing the framework can declare whether it is supporting HPOS. We could accomplish this by adding a property SV_WC_Plugin::$supports_hpos : bool (default true) and have individual plugins override it if necessary (plugins adopting the new Framework version should support HPOS by default).

## QA

- [ ] Code review
- [ ] Override the `hpos_support`  value to `false` in the gateway class
- [ ] You will see this warning in WooCommerce > Settings > Advanced > Features.
image

### DIFF
--- a/woocommerce/class-sv-wc-plugin.php
+++ b/woocommerce/class-sv-wc-plugin.php
@@ -45,6 +45,9 @@ abstract class SV_WC_Plugin {
 	/** Plugin Framework Version */
 	const VERSION = '5.10.14';
 
+	/** Default support for HPOS */
+	const SUPPORTS_HPOS = true;
+
 	/** @var object single instance of plugin */
 	protected static $instance;
 
@@ -275,6 +278,9 @@ abstract class SV_WC_Plugin {
 
 		// hook for translations separately to ensure they're loaded
 		add_action( 'init', array( $this, 'load_translations' ) );
+
+		// Add high performance order storage compatibility
+		add_action( 'before_woocommerce_init', array( $this, 'add_hpos_compatibility' ) );
 
 		// add the admin notices
 		add_action( 'admin_notices', array( $this, 'add_admin_notices' ) );
@@ -1454,6 +1460,17 @@ abstract class SV_WC_Plugin {
 		wc_deprecated_function( __METHOD__, '5.2.0' );
 	}
 
+	/**
+	 * Add the HPOS compatibility.
+	 *
+	 * @since x.y.z
+	 */
+	public function add_hpos_compatibility() {
+
+		if ( class_exists( \Automattic\WooCommerce\Utilities\FeaturesUtil::class ) ) {
+			\Automattic\WooCommerce\Utilities\FeaturesUtil::declare_compatibility( 'custom_order_tables', $this->get_plugin_file(), static::SUPPORTS_HPOS );
+		}
+	}
 
 }
 
--- a/woocommerce/class-sv-wc-plugin.php
+++ b/woocommerce/class-sv-wc-plugin.php
@@ -63,6 +63,9 @@ abstract class SV_WC_Plugin {
 	/** @var string template path, without trailing slash */
 	private $template_path;
 
+	/** @var bool whether the plugin supports WooCommerce HPOS */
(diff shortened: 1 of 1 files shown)


woocommerce/class-sv-wc-plugin.php
```diff
@@ -1454,6 +1460,17 @@ protected function set_dependencies( $dependencies = [] ) {
 		wc_deprecated_function( __METHOD__, '5.2.0' );
 	}
 
+	/**
+	 * Add the HPOS compatibility.
+	 *
```
unfulvio-godaddy, line 1465: mark as `@internal`

woocommerce/class-sv-wc-plugin.php
```diff
@@ -45,6 +45,9 @@ abstract class SV_WC_Plugin {
 	/** Plugin Framework Version */
 	const VERSION = '5.10.14';
 
+	/** Default support for HPOS */
+	const SUPPORTS_HPOS = true;
```
unfulvio-godaddy, line 49: I would use a protected property for this

unfulvio-godaddy: it would be in line with plugins setting arguments from their constructor - that's when they could set this internal flag

unfulvio-godaddy: the parent constructor can apply properties passed from args as it does currently

unfulvio-godaddy (changes requested):

woocommerce/class-sv-wc-plugin.php
jstern-godaddy: I think we should put an explicit notice in the changelog warning developers to be sure to set this flag to `false` if they're using this version of the framework and _haven't_ yet implemented HPOS compatibility

woocommerce/class-sv-wc-plugin.php
```diff
@@ -1454,6 +1460,17 @@ protected function set_dependencies( $dependencies = [] ) {
 		wc_deprecated_function( __METHOD__, '5.2.0' );
 	}
 
+	/**
+	 * Add the HPOS compatibility.
+	 *
+	 * @since x.y.z
+	 */
+	public function add_hpos_compatibility() {
+
```
unfulvio-godaddy, line 1469: because of this bug https://godaddy.slack.com/archives/C026JPZ4GLB/p1673409266503829 it's quite possible that we will need additional logic in this method - for instance check if the installed WC version is the one that contains the bugfix

woocommerce/class-sv-wc-plugin.php
```diff
@@ -1454,6 +1460,17 @@ protected function set_dependencies( $dependencies = [] ) {
 		wc_deprecated_function( __METHOD__, '5.2.0' );
 	}
 
+	/**
+	 * Add the HPOS compatibility.
+	 *
+	 * @since x.y.z
+	 */
+	public function add_hpos_compatibility() {
```
unfulvio-godaddy, line 1468:
```suggestion
	public function handle_hpos_compatibility() {
```

it may not add it since it's conditional

woocommerce/class-sv-wc-plugin.php
unfulvio-godaddy: or perhaps we can be more conservative about it and default to `false` - but either way that could have unintended effects if the devs are not sure how to implement that... I'm actually leaning to default `false`

sharazghouri: @unfulvio-godaddy I agree with your opinion. we should use the `protected` property with a default value is `false`.

sharazghouri: Because constant always priests in memory.

sharazghouri: @unfulvio-godaddy My intention to make it constant is that somehow it helps us to detect whether the weather plugin supports HPOS.

shehrozsheikh (approved):

unfulvio-godaddy (approved):

PR #583, opened 2023-01-11
Added static helper methods to detect admin screen_id and get orders

shehrozsheikh:
# Summary
This PR contains adds helper methods for HPOS-related functionalities.

### Details
- Added a helper method to detect the admin screen_id
- Added a helper method to replace the `get_post_type` method with the new logic for order detection, and replaced all the usages in the FW
- ~~Added a helper method to use get WC Orders using the new query arguments introduced with HPOS.~~

### DIFF
--- a/woocommerce/compatibility/class-sv-wc-order-compatibility.php
+++ b/woocommerce/compatibility/class-sv-wc-order-compatibility.php
@@ -569,12 +569,11 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
 	 * @since x.y.z
 	 *
 	 * @param array $args Array of args.
-	 * @param bool $use_get_posts. Whether to use get_posts() method if HPOS is disabled. Defaults to false.
 	 * @return mixed
 	 */
-	public static function get_orders( $args, bool $use_get_posts = false ) {
+	public static function get_orders( $args ) {
 
-		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() && $use_get_posts ) {
+		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
 			return get_posts( $args );
 		}
 
--- a/woocommerce/compatibility/class-sv-wc-order-compatibility.php
+++ b/woocommerce/compatibility/class-sv-wc-order-compatibility.php
@@ -534,7 +534,7 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
 	 *
 	 * @return string
 	 */
-	public static function get_screen_id( ) {
+	public static function get_screen_id( ) : string {
 
 		return SV_WC_Plugin_Compatibility::is_hpos_enabled() ? wc_get_page_screen_id( 'shop-order' ) : 'shop_order';
 	}
@@ -548,7 +548,7 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
 	 * @param int $post_id post ID.
 	 * @return bool
 	 */
-	public static function is_order( $post_id ) {
+	public static function is_order( int $post_id ) : bool {
 
 		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
 			return 'shop_order' === get_post_type( $post_id );
--- a/woocommerce/compatibility/class-sv-wc-order-compatibility.php
+++ b/woocommerce/compatibility/class-sv-wc-order-compatibility.php
@@ -24,6 +24,8 @@
 
 namespace SkyVerge\WooCommerce\PluginFramework\v5_10_14;
 
+use Automattic\WooCommerce\Utilities\OrderUtil;
+
 defined( 'ABSPATH' ) or exit;
 
 if ( ! class_exists( '\\SkyVerge\\WooCommerce\\PluginFramework\\v5_10_14\\SV_WC_Order_Compatibility' ) ) :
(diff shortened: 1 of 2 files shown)


woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,64 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+
+	/**
+	 * Determines whether the given $post_id is a WooCommerce order or not according to HPOS availability.
+	 *
+	 * @since x.y.z
+	 *
+	 * @param int $post_id post ID.
+	 * @return bool
+	 */
+	public static function is_order( $post_id ) {
```
unfulvio-godaddy, line 551:
```suggestion
	public static function is_order_screen( int $post_id ) : bool {
```

I recommend changing this to `is_order_screen` to avoid confusions with the order object

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,64 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * This method detects the screen_id according to HPOS availability,
+	 * as cannot rely anymore on the shop_order registered post type as the screen ID.
+	 *
+	 * @since x.y.z
+	 *
+	 * @return string
+	 */
+	public static function get_screen_id( ) {
+
+		return SV_WC_Plugin_Compatibility::is_hpos_enabled() ? wc_get_page_screen_id( 'shop-order' ) : 'shop_order';
```
unfulvio-godaddy, line 539: this method is pending another story that will add this method: should that story/pr added first?

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,64 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * Determines whether the given $post_id is a WooCommerce order or not according to HPOS availability.
+	 *
+	 * @since x.y.z
+	 *
+	 * @param int $post_id post ID.
+	 * @return bool
+	 */
+	public static function is_order( $post_id ) {
+
+		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
```
unfulvio-godaddy, line 553: same here, pending for this method to be added

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,64 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * @return mixed
+	 */
+	public static function get_orders( $args, bool $use_get_posts = false ) {
+
+		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() && $use_get_posts ) {
+			return get_posts( $args );
+		}
+
+		return wc_get_orders( $args );
+	}
```
unfulvio-godaddy, line 582: adding `$use_get_posts` gives the burden of determining which arguments to use and are compatible with `wc_get_order` on each plugin - besides, arguments passed to get_posts aren't all/always interchangeable with wc_get_orders and thus this may introduce bugs

probably this compatibility method needs to adapt arguments coming from get_orders and adapt them to work with either wc_get_orders or get_posts whether or not hpos is enabled

unfulvio-godaddy (changes requested):

woocommerce/compatibility/class-sv-wc-order-compatibility.php
shehrozsheikh: @unfulvio yes, it's related to [MWC-10118](https://godaddy-corp.atlassian.net/browse/MWC-10118), so I added another story here.

It's coming from the 'Methods to upgrade' section in the [Discovery Doc](https://docs.google.com/document/d/1752eyK4-9d-bxlOc2ZbAbZqEk91O5aPPrZHC-U0ovYE/edit#).

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: Hey @shehrozsheikh to further clarify:

Suppose we have a plugin using `get_posts()` to fetch some `shop_order` posts. With HPOS that won’t work anyore. So, whenever possible we would just rewrite that query using `wc_get_orders()`, ideally. So far so good. 

But we might encounter a few problems. 

* Arguments passed to `get_posts()` aren’t directly interchangeable with wc_get_orders(): https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query
* In particular handling meta it’s very different: https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#adding-custom-parameter-support
* With the new WC version introducing HPOS we have yet new arguments that make this a bit easier: https://github.com/woocommerce/woocommerce/wiki/HPOS:-new-order-querying-APIs and closer to get_posts()

So, we have to account for the possible scenarios:

1. Installation not using HPOS, using an older WC version (use get_posts or wc_get_orders with the arguments available for this combination)
2. Installation not using HPOS, using the latest WC version (but it’s not clear if the arguments introduced in https://github.com/woocommerce/woocommerce/wiki/HPOS:-new-order-querying-APIs will also work in this scenario - but looks like not - please confirm if that's the case)
3. Installation using HPOS, using the latest WC version (then we sure know those args are available and we can use them)

I believe regardless of case n.2 above, it seems we will have 2 paths.

1. We could handle the compatibility change at plugins level for better clarity at the cost of code duplication (ie introduce a if statement when `get_posts()` is used and _cannot_ be simply moved to `wc_get_orders()` because it queries by order meta or does something the legacy `wc_get_orders()` can't do in a straightforward way. 
2. Another way is to introduce a compatibility method in the framework that parses the arguments of a query _assuming the latest `wc_get_orders()`_ and maps those arguments to `get_posts()` for legacy or passes them to `wc_get_orders()` proper for HPOS. The day we can drop support for legacy order posts this will be easier to transition because all the plugins will be using arguments compatible with `wc_get_orders()` with HPOS. 

I'll bring this up to the channel since it is somewhat complex, but also to make sure we're on the same page.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,64 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+
+	/**
+	 * Determines whether the given $post_id is a WooCommerce order or not according to HPOS availability.
+	 *
+	 * @since x.y.z
+	 *
+	 * @param int $post_id post ID.
+	 * @return bool
+	 */
+	public static function is_order( int $post_id ) : bool {
```
unfulvio-godaddy, line 551: consider renaming method to is_order_admin_screen() : bool to avoid confusion with determining if a a post ID is an order; additionally, you could make this method to work also without passing a $post_id (make the post_id arg nullable) by just checking if the current page is an order admin screen

woocommerce/compatibility/class-sv-wc-order-compatibility.php
shehrozsheikh: Thanks, @unfulvio-godaddy for putting the problems and possible solutions altogether.

I also think that mapping may cause some issues, so this logic must be handled inside a plugin.

I have updated the method and added some detailed descriptions in phpdocs, feel free to update it.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: my bad I misinterpreted this

unfulvio-godaddy (approved):
@shehrozsheikh I have removed the compatibility method for get_orders considering our prior discussion about this - maybe it's best if individual plugins try to convert instances of get_posts() to wc_get_orders() (which is compatible also with older versions of WC) - when it comes to meta data they need to use either get_posts or wc_get_orders() using the new meta queries available after HPOS

PR #584, opened 2023-01-12
HPOS Compatibility

### LINKED ISSUE #570: Use `set_transaction_id` method of WC_Order to set transaction ID.
This is an enhancement request. Framework uses `update_post_meta` to set transaction id to order, and it can create an issue for the upcoming "Custom Order Table" functionality of WooCommerce.
https://github.com/godaddy-wordpress/wc-plugin-framework/blob/master/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php#L2517

[set_transaction_id](https://woocommerce.github.io/code-reference/classes/WC-Order.html#method_set_transaction_id) method should be used here as a more future-proof solution. 

Example:
```php
$order->set_transaction_id( $response->get_transaction_id() );
$order->save();
```

Thanks

### LINKED ISSUE #590: Return type of SV_WC_DateTime::getTimestamp()
wc-plugin-framework/woocommerce/compatibility/class-sv-wc-datetime.php:86

Return type of SV_WC_DateTime::getTimestamp() should either be compatible with DateTime::getTimestamp(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice.

This is a bug that anyone on php8+ experiences in gateways that use this framework. The solution which needs to be added to the code, is to change from:

`public function getTimestamp() {`

to

`public function getTimestamp():int {`

unfulvio-godaddy:
## Summary

This is the base branch which will contain all changes related to HPOS compatibility.

### Epic [MWC-10156](https://godaddy-corp.atlassian.net/browse/MWC-10156)

This PR comes with a set of helpers for orders in `SV_WC_Plugin_Compatibility` and `SV_WC_Order_Compatibility` to handle order objects conditionally whether the installation uses HPOS or not.

To declare a plugin adopting this version of the FW as HPOS ready, add the following argument in the plugin main class constructor that calls the parent `SV_WC_Plugin` constructor:

```php
			'supports_hpos' => true,
```

Make sure you also update the plugin `composer.json` file to point the Framework to this branch of the repository:

```json
    "skyverge/wc-plugin-framework": "dev-hpos-support as 5.11.0",
```

Then update composer with:

```shell
composer update skyverge/wc-plugin-framework --no-dev
```

Finally, remember to bump the main plugin file `FRAMEWORK_VERSION` constant to `5.11.0` and all the framework namespaces referenced in the plugin with `5_11_0` to match the version.

The plugin should run this version of the FW and declare itself compatible with HPOS. _However_ you may still need to add plugin-specific compatibility code and do the necessary updates, possibly using the helper methods provided in this version of the FW.

Because of  [this WC bug](https://github.com/woocommerce/woocommerce/issues/35703) the minimum required version to use HPOS with a frameworked plugin is WC 7.6

#### Fixes #570


woocommerce/class-sv-wc-plugin.php
```diff
@@ -790,6 +814,19 @@ public function is_tls_1_2_available() {
...
+	/**
+	 * Determines whether the plugin supports HPOS.
+	 *
+	 * @since x.y.z
+	 *
+	 * @return bool
+	 */
+	public function is_hpos_compatible() : bool
+	{
+		return $this->supports_hpos;
```
unfulvio-godaddy, line 826: TODO: we might need to add a WC version check here for when this bug has been addressed: https://github.com/woocommerce/woocommerce/issues/35703

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,182 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+
+			if ( $post_order_or_id instanceof \WC_Order ) {
+				$post_order_or_id = $post_order_or_id->get_id();
+			}
+
+			return is_numeric( $post_order_or_id ) || $post_order_or_id instanceof \WP_Post
+				&& 'shop_order' === get_post_type( $post_order_or_id );
+		}
+
+		return 'shop_order' === OrderUtil::get_order_type( $post_order_or_id );
```
unfulvio-godaddy, line 565: consider that refunds and subscriptions may not work here - when handling plugins we can check if additional helper methods are required for detecting those specifically

unfulvio-godaddy: Consider

```php
OrderUtil::is_order( $post_id, wc_get_order_types() );
```

this would detect any order type

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,200 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * @param string $meta_key meta key
+	 * @param bool $single return the first found meta with key (true), or all meta sharing the same key (default true)
+	 * @return mixed
+	 */
+	public static function get_order_meta( $order, string $meta_key, bool $single = true ) {
+
+		if ( SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
+
+			$value = $single ? '' : [];
+			$order = ! $order instanceof \WC_Order && is_numeric( $order ) ? wc_get_order( (int) $order ) : $order;
```
jstern-godaddy, line 604: @sahmed2-godaddy just wondering if there's a reason why this conditional includes both `! $order instanceof \WC_Order` and `is_numeric( $order )` ? If `is_numeric( $order )` is true, then by definition `! $order instanceof \WC_Order` is also true.

Why not just check `is_numeric( $order )` and return `wc_get_order( (int) $order )` if true, otherwise just returning `$order`?

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -522,6 +524,200 @@ public static function get_edit_order_url( \WC_Order $order ) {
...
+	 * Gets the admin screen ID for orders.
+	 *
+	 * This method detects the expected orders screen ID according to HPOS availability.
+	 * `shop_order` as a registered post type as the screen ID is no longer used when HPOS is active.
+	 *
+	 * @since x.y.z
+	 *
+	 * @return string
+	 */
+	public static function get_screen_id( ) : string {
```
sshadid-godaddy, line 537: The name should be `get_admin_order_screen_id()` or `get_order_screen_id()` for better readability.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
sahmed2-godaddy: @jstern-godaddy agreed, @unfulvio-godaddy made this change in a recent commit so unless he has a specific reason behind this change, I will update it.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: Aren't screens only for admin contexts in WP? I'd vote for the latter.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
unfulvio-godaddy: agreed it's redudant

unfulvio-godaddy:
fixes #570

jstern-godaddy (approved):

PR #585, opened 2023-01-17
Update references to WP meta functions to use WC_Data CRUD equivalents

unfulvio-godaddy:
## Summary

This PR updates instances where the FW tries to handle order post meta using WordPress functions like add/get/update/delete_post_meta instead of the WC_Data equivalents. Same applies to Subscriptions: we should use WC methods instead of WP meta functions when handling Subscriptions meta data.

## QA

- [ ] Code review
- [ ] Audit the FW codebase [using this regex](https://github.com/woocommerce/woocommerce/wiki/High-Performance-Order-Storage-Upgrade-Recipe-Book#auditing-the-code-base-for-direct-db-access-usage): you shouldn't see any more instances of WP meta functions used in the codebase with this PR


agibson-godaddy (approved):
LGTM

PR #587, opened 2023-01-30
Update admin_footer-post.php and load-edit.php hooks for handling HPOS screens

sharazghouri:
## Summary

While testing Authorize mode transaction, I noticed that the Capture Charge button is not showing on the order edit screen.  I figured out why it's not showing,   [admin_footer-edit.php](https://github.com/godaddy-wordpress/wc-plugin-framework/blob/master/woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php#L65) hook not working in HPOS-supported WooCommecre.  First, it's not a normal post-type edit screen that's why this hook not working.

### DIFF
--- a/woocommerce/compatibility/class-sv-wc-order-compatibility.php
+++ b/woocommerce/compatibility/class-sv-wc-order-compatibility.php
@@ -722,42 +722,6 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
 	}
 
 
-	/**
-	 * Get global $post or $theorder order object according to HPOS availability.
-	 *
-	 * @since x.y.z
-	 * @return \WC_Order|\WP_Post  object
-	 */
-	public static function get_post_or_order() {
-
-		global $post, $theorder;
-
-		if ( ! is_null( $theorder ) && static::is_order( $theorder ) ) {
-			return $theorder;
-		} else if( ! is_null( $post ) ) {
-			return $post;
-		}
-
-	}
-
-	/**
-	 * Get global $post or $theorder order status.
-	 *
-	 * TODO: Refactor this method to get order status by order object or ID. If object order or ID does not pass in params then get global order status.
-	 * @since x.y.z
-	 * @return string order status
-	 */
-	public static function get_order_status() {
-
-	$post_or_order_object	= static::get_post_or_order();
-
-		if ( $post_or_order_object instanceof \WC_Order ) {
-			return $post_or_order_object->get_status();
-		} elseif ( $post_or_order_object instanceof \WP_Post ) {
-			return get_post_status( $post_or_order_object );
-		}
-	}
-
 }
 
 
--- a/woocommerce/compatibility/class-sv-wc-order-compatibility.php
+++ b/woocommerce/compatibility/class-sv-wc-order-compatibility.php
@@ -641,11 +641,11 @@ class SV_WC_Order_Compatibility extends SV_WC_Data_Compatibility {
 			return false;
 		}
 
-		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
+		if ( $post_order_or_id instanceof \WC_Abstract_Order ) {
 
-			if ( $post_order_or_id instanceof \WC_Order ) {
-				$post_order_or_id = $post_order_or_id->get_id();
-			}
+			$found_type = $post_order_or_id->get_type();
+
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/admin/class-sv-wc-payment-gateway-admin-order.php
```diff
@@ -137,9 +137,11 @@ protected function enqueue_edit_order_assets( \WC_Order $order ) {
...
 	public function maybe_add_capture_charge_bulk_order_action() {
-		global $post_type, $post_status;
+		
+		global $post_type, $post_status, $current_screen;
 
-		if ( ! current_user_can( 'edit_shop_orders' ) ) {
+
+		if ( ! current_user_can( 'edit_shop_orders' )  && $current_screen->id === SV_WC_Order_Compatibility::get_screen_id()) {
 			return;
 		}
```
sharazghouri, line 147: `$post_type` and `$post_status` will not work in WooCommerce to check post type and post status. We need more helper methods for these operations.  I'm working on it.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -718,6 +718,42 @@ public static function delete_order_meta( $order, string $meta_key, $meta_value
...
+
+		global $post, $theorder;
+
+		if ( ! is_null( $theorder ) && static::is_order( $theorder ) ) {
+			return $theorder;
+		} else if( ! is_null( $post ) ) {
+			return $post;
+		}
+	
+	}
```
unfulvio-godaddy, line 737: I don't think this method would be needed. As it is currently it also returns void / does not handle scenarios where global is neither order or post. But besides, I don't think we want methods that return different objects that aren't related.

woocommerce/compatibility/class-sv-wc-order-compatibility.php
```diff
@@ -718,6 +718,42 @@ public static function delete_order_meta( $order, string $meta_key, $meta_value
...
+
+	$post_or_order_object	= static::get_post_or_order();
+
+		if ( $post_or_order_object instanceof \WC_Order ) {
+			return $post_or_order_object->get_status();
+		} elseif ( $post_or_order_object instanceof \WP_Post ) {
+			return get_post_status( $post_or_order_object );
+		}
+	}
+
```
unfulvio-godaddy, line 756: I am also not sure if this method is needed either.

Also this method does not account for the possibility neither global is set and will return void.

But also, I believe if the purpose of adding this is solely to fix the logic in the hook callback below, then we could just move the corresponding logic in the callback. I made a quick search on plugins and I don't think there are more instances of `admin_footer-edit.php`. Sure enough, there could be more contexts where we need detection of an order edit screen - but in that case we could benefit from a different helper method.

unfulvio-godaddy (changes requested):

unfulvio-godaddy (commented):
I checked the results of get_current_screen and other data available in the orders screen or the order edit screen... it's not ideal but maybe we can just use both the screen id and elements from the `$_GET` request until we find a better/robust way to handle this.

Something along the lines of:

```php


	/**
	 * Determines if the current admin screen is for the orders.
	 * 
	 * @return bool
	 */
	public static function is_orders_screen() : bool {

		$current_screen = SV_WC_Helper::get_current_screen();

		if ( ! $current_screen ) {
			return false;
		}

		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
			return 'edit-shop_order' === $current_screen->id;
		}

		return static::get_order_screen_id() === $current_screen->id
			&& isset( $_GET['page'] )
			&& $_GET['page'] === 'wc-orders';
	}

	
	/**
	 * Determines if the current admin screen is for adding or editing an order. 
	 * 
	 * @since x.y.z
	 * 
	 * @return bool
	 */
	public static function is_order_edit_screen() : bool {

		$current_screen = SV_WC_Helper::get_current_screen();

		if ( ! $current_screen ) {
			return false;
		}

		if ( ! SV_WC_Plugin_Compatibility::is_hpos_enabled() ) {
			return 'edit-shop_order' === $current_screen->id;
		}

		return static::get_order_screen_id() === $current_screen->id
			&& isset( $_GET['page'], $_GET['action'] )
			&& $_GET['page'] === 'wc-orders'
			&& in_array( $_GET['action'], [ 'new', 'edit'], true );
	}


```

unfulvio-godaddy (approved):

PR #594, opened 2023-03-25
Handle empty array for gateway card types

nikolas4175-godaddy:
# Summary
Resolves an issue where customers would encounter critical errors when loading the payment form at checkout if a merchant had no card types selected in the gateway's "Accepted Card Logos" setting.

### Setup

- Install and activate a gateway plugin that supports the "Card Types" feature (e.g. ANet)

### Steps

1. Remove all card types from the gateway's "Accepted Card Logos" setting. Save changes.
1. Proceed to checkout 
	- [ ] The payment form loads on checkout (with no card logos) without error


woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -3366,7 +3366,7 @@ public function get_card_types() {
 
 		assert( $this->supports_card_types() );
 
-		return $this->card_types;
+		return is_array($this->card_types) ? $this->card_types : array();
```
unfulvio-godaddy, line 3369:
```suggestion
		return is_array( $this->card_types ) ? $this->card_types : [];
```

unfulvio-godaddy (approved):

PR #598, opened 2023-05-29
Release: 5.11.2 (Remove AllowDynamicProperties at method level)

unfulvio-godaddy:
Unfortunately we slipped a `#[\AllowDynamicProperties]` at a method level during the release of 5.11.0 - which not only isn't helpful, but although assumed harmless, it actually causes a fatal error in PHP 8.2. 

Since the object where we would like to add some dynamic properties is the `WC_Order` object, this has no effect anyway, even at the class level which extends `WC_Payment_Gateway`. 

This currently is only affecting Authorize.Net running PHP 8.2 and we should fix that by updating the FW.

Slack discussion: https://godaddy.slack.com/archives/C04L7S7JJE9/p1685116298433759
Affected plugin PR: https://github.com/gdcorp-partners/woocommerce-gateway-authorize-net-cim/pull/84


pbozionelos-godaddy:
Hey @unfulvio-godaddy . I was able to test this using MAMP and PHP 8.2. I was able to replicate the issue with the previous issue of the plugin and when I changed to this PR and installed the latest version of the framework, the issue seems to be resolved. Not sure if this should be affecting anything else, but I'm happy to do any more tests if you'd like.

unfulvio-godaddy:
thanks @pbozionelos-godaddy sounds good to me

PR #610, opened 2023-07-26
Enqueue form styles for hosted payment forms

ragulka:
# Summary

This PR does 2 things:

1. Ensure FW payment form styles are enqueued for gateways that use embedded / hosted payment forms.
2. Update payment selector styles in an attempt to always align the gateway icon and CC icons to the right. 

The latter change is bigger of the 2, and affects all FW-d plugins. I tested with 3 different gateways on different scenarios. I'm attaching the screenshots of the results. Ideally, we would contribute to WC core an update to use flexbox for the selector styling, as this makes achieving the proper alignment much easier.

CleanShot 2023-07-26 at 15 58 15@2x
CleanShot 2023-07-26 at 15 57 32@2x
CleanShot 2023-07-26 at 15 57 05@2x


unfulvio-godaddy (approved):
LGTM - I have added a changelog entry, and a since tag to the newly added method

PR #614, opened 2023-09-21
Release: v5.12.0 -  Add Cart/Checkout Blocks support

unfulvio-godaddy:
## Summary

This PR adds support for WooCommerce Cart/Checkout blocks on the backend side.

### Main changes required by plugins switching to this version:

- Plugins should declare their cart/checkout support by passing `$args['supported_features']['blocks']['checkout'] = true` in their main plugin constructor arguments (this was already backported into into released v5.11.12)
- Non-gateway plugins should include a class that extends a `Blocks_Handler` class, similar as they do with the `Lifecycle_Handler`; this class is responsible for initializing any integration classes
- Gateway plugins might do the same but extending a `Gateway_Blocks_Handler` instead; however, their individual gateways may just implement  `get_cart/checkout_block_integration_instance` methods to provide an instance of the block integration they want to declare, and that will be taken care automatically
- The reason for this difference is that gateway block integrations may need to extend `Checkout_Block_Integration` which also requires passing a gateway instance in the constructor - also, the way block integrations for non-payment methods register is different, and a payment gateway could have different methods that need to implement their own integration (credit card, echeck, etc)

A small note on external checkouts that extend `External_Checkout` class. These classes may now pass a `supported_features` property in optional arguments in their constructor to define if they also support either block, in a similar way as the main plugin file does. The fact we have this duplication allows for a plugin to approach compatibility in iterations, and perhaps sort with its gateway components first and then external checkout later. If not indicated, it will default to the general plugin compatibility values.

### Payment Method Payload

Please [check this document](https://docs.google.com/document/d/1jjFnsWxSXtpjj_Exkpv-CVblDMbBf7Un7pQO8h45Wag/edit) for an overview of the payment method data payload handled by the FW.

Individual gateways may add additional data as required.

### Expected assets to be enqueued

The CSS/JS file to be included by plugins adding front end support is by default built as follows:

* CSS: `wc-<plugin-name>-<block-name>-block.css`
* JS `wc-<plugin-name>-<block-name>-block.js`

For example `wc-authorize-net-cim-checkout-block.css`

For dev purposes if you want to point those assets to alternative URLs, you can use the following hooks:

* CSS: `'wc_' . $this->get_id() . '_' . $this->block_name . '_block_stylesheet_url'`
* JS: ` 'wc_' . $this->get_id() . '_' . $this->block_name . '_block_script_url'`

where the id would correspond to the plugin in context e.g. `wc_authorize_net_cim_checkout_script_url` (we are going with the assumption that plugins with multiple gateways will use a single script/stylesheet for all their gateways)


woocommerce/payment-gateway/Blocks/Gateway_Checkout_Block_Integration.php
```diff
@@ -114,8 +114,7 @@ public function get_payment_method_script_handles() {
 	 *
 	 * @return array<string, mixed>
 	 */
-	public function get_script_data()
-	{
+	public function get_payment_method_data() : array {
```
itambek-godaddy, line 116: @unfulvio-godaddy this method seems to be required for payment gateways so their settings are available via `getSetting` in frontend.

woocommerce/Blocks/Traits/Block_Integration_Trait.php
```diff
@@ -36,7 +36,7 @@ trait Block_Integration_Trait
 	 */
 	public function get_name() {
 
-		return $this->plugin->get_id_dasherized();
+		return $this->plugin->get_id();
```
itambek-godaddy, line 39: @unfulvio-godaddy looks like Woo's convention is to use underscored names, for example `WC_Gateway_Braintree_Credit_Card_Blocks_Support` sets the name as `braintree_credit_card`

PR #625, opened 2023-10-06
Address translator feedback and add context to gettext

unfulvio-godaddy:
This PR addresses the issues reported in: https://godaddy.slack.com/archives/C07TDQWJ1/p1696533437732739?thread_ts=1695360403.201719&cid=C07TDQWJ1


woocommerce/Lifecycle.php
```diff
@@ -390,18 +390,18 @@ protected function generate_milestone_notice_message( $custom_message ) {
...
+			$exclamations = [
+				_x( 'Awesome!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+				_x( 'Fantastic!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+				_x( 'Cowabunga!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+				_x( 'Congratulations!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+				_x( 'Hot dog!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+			];
 
-			$message = $exclamations[ array_rand( $exclamations ) ] . ', ' . esc_html( $custom_message ) . ' ';
+			$message = $exclamations[ array_rand( $exclamations ) ] . ' ' . esc_html( $custom_message ) . ' ';
```
unfulvio-godaddy, line 401: note: I have adjusted this because in some languages the comma may create a problem, I replaced it with an exclamation, ie. starting a new sentence. This probably is still not bullet-proof for RTL contexts but it's better than before

woocommerce/class-sv-wc-plugin.php
```diff
@@ -716,7 +716,7 @@ public function add_system_status_php_information( $rows ) {
 					continue;
 				}
 
-				/* translators: As in "Value has been set as [foo], but [bar] is required". Placeholders: %1$s - current value for a PHP setting, %2$s - required value for the PHP setting */
+				/* translators: Context: As in "Value has been set as [foo], but [bar] is required". Placeholders: %1$s - current value for a PHP setting, %2$s - required value for the PHP setting */
 				$note = __( 'Set as %1$s - %2$s is required.', 'woocommerce-plugin-framework' );
```
unfulvio-godaddy, line 720: the translator feedback said they didn't understand this line and to provide examples, but it's possible they were looking at an outdated file that didn't have the translators comment

woocommerce/payment-gateway/External_Checkout/Admin.php
```diff
@@ -390,13 +390,8 @@ protected function get_configuration_errors() {
...
 
 			$errors[] = sprintf(
-				/* translators: Placeholders: %1$s - plugin name, %2$s - a currency/comma-separated list of currencies, %3$s - <a> tag, %4$s - </a> tag, %5$s - external checkout label */
-				_n(
-					'Accepts payment in %1$s only. %2$sConfigure%3$s WooCommerce to accept %1$s to enable %4$s.',
-					'Accepts payment in one of %1$s only. %2$sConfigure%3$s WooCommerce to accept one of %1$s to enable %4$s.',
-					count( $accepted_currencies ),
-					'woocommerce-plugin-framework'
-				),
+				/* translators: Context: Error message displayed to merchants if their store currency is not in one of the accepted currencies by the payment gateway they intend to enable. Placeholders: %1$s - a currency or comma-separated list of currencies, %2$s - opening HTML <a> tag, %3$s - closing HTML </a> tag, %4$s - payment method title/label */
```
unfulvio-godaddy, line 393: the problem here was that the translators placeholder was placed above the line where the translatable text was, maybe it wasn't parsed correctly in gettext, so in this occurrence, and others, I have rearranged it

woocommerce/payment-gateway/External_Checkout/Google_Pay/Admin.php
```diff
@@ -99,7 +99,7 @@ public function get_settings() {
 			[
 				'id'      => 'sv_wc_google_pay_display_locations',
 				/* translators: Allow Google Pay button on selected display locations (e.g. cart, checkout, product page...) */
-				'title'   => __( 'Allow Google Pay on', 'woocommerce-plugin-framework' ),
+				'title'   => _x( 'Allow Google Pay on', 'Setting title for a multi-select list dropdown of display locations', 'woocommerce-plugin-framework' ),
```
unfulvio-godaddy, line 102: the translator didn't understand the context, but it's possible that they were looking at a file that was missing the translators comment

woocommerce/payment-gateway/External_Checkout/Admin.php
```diff
@@ -429,14 +424,13 @@ protected function should_display_shipping_based_tax_notice() {
...
 	 * @return string
 	 */
-	protected function get_shipping_based_tax_notice() {
+	protected function get_shipping_based_tax_notice() : string {
 
 		return sprintf(
-			/** translators: Placeholders: %1$s - external checkout label, %2$s - <a> tag, %3$s - </a> tag, %4$s - <strong> tag, %5$s - </strong> tag */
-			__( '%4$s%1$s Notice!%5$s Your store %2$scalculates taxes%3$s based on the shipping address, but %1$s %4$sdoes not%5$s share customer shipping information with your store for orders with only virtual products. These orders will have their taxes calculated based on the shop address instead.', 'woocommerce-plugin-framework' ),
+			/* translators: Placeholders: %1$s - Checkout method label, %2$s - Opening HTML <a> tag, %3$s - Closing HTML </a> tag, %4$s - Opening HTML <strong> tag, %5$s - closing HTML </strong> tag */
+			__( 'Your store %1$scalculates taxes%2$s based on the shipping address, but %1$s %4$sdoes not%5$s share customer shipping information with your store for orders with only virtual products. These orders will have their taxes calculated based on the shop address instead.', 'woocommerce-plugin-framework' ),
```
unfulvio-godaddy, line 431: I changed slightly this notice as it was hard to parse by the translator - basically I just trimmed the "Notice!" bit which I too found a bit weirdly worded

woocommerce/Lifecycle.php
```diff
@@ -390,18 +390,18 @@ protected function generate_milestone_notice_message( $custom_message ) {
...
-				__( 'Awesome', 'woocommerce-plugin-framework' ),
-				__( 'Fantastic', 'woocommerce-plugin-framework' ),
-				__( 'Cowabunga', 'woocommerce-plugin-framework' ),
-				__( 'Congratulations', 'woocommerce-plugin-framework' ),
-				__( 'Hot dog', 'woocommerce-plugin-framework' ),
-			);
+			$exclamations = [
+				_x( 'Awesome!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+				_x( 'Fantastic!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
+				_x( 'Cowabunga!', 'Generic congratulatory exclamation at the start of a milestone message', 'woocommerce-plugin-framework' ),
```
itambek-godaddy, line 396: 😂

woocommerce/payment-gateway/External_Checkout/Admin.php
itambek-godaddy: Yup, that was probably it

woocommerce/payment-gateway/class-sv-wc-payment-gateway-direct.php
```diff
@@ -746,21 +747,21 @@ protected function do_credit_card_transaction( $order, $response = null ) {
...
+					$this->is_test_environment() ? esc_html_x( 'Test', 'Noun, software environment', 'woocommerce-plugin-framework' ) : '',
+					$this->perform_credit_card_authorization( $order ) ? esc_html_x( 'Authorization', 'Credit card transaction type', 'woocommerce-plugin-framework' ) : esc_html_x( 'Charge', 'noun, credit card transaction type', 'woocommerce-plugin-framework' ),
 					SV_WC_Payment_Gateway_Helper::payment_type_to_name( $card_type ),
 					$last_four
 				);
 			} else {
 				$message = sprintf(
-					/* translators: Placeholders: %1$s - payment method title, %2$s - environment ("Test"), %3$s - transaction type (authorization/charge), %4$s - card type (mastercard, visa, ...) */
+					/* translators: Example: "Authorize.Net Test Charge Approved: Mastercard" - Placeholders: %1$s - payment method title, %2$s - environment ("Test"), %3$s - transaction type (authorization/charge), %4$s - card type (mastercard, visa, ...) */
 					esc_html__( '%1$s %2$s %3$s Approved: %4$s', 'woocommerce-plugin-framework' ),
```
itambek-godaddy, line 761: Not a comment on the current changes, but something like this is extremely difficult to translate, because the placeholders are also translatable and may ned to use different inflections based on the context or the grammatical case used...

For example, in Estonian this sentece will become something like:
```
Authorize.net Testkeskkond Kaardimakse vastu võetud: Mastercard
```
But it's wrong, it should be (no capitalization, different suffix/inflection for "testkeskkond") - same [as here](https://x.com/paragulka/status/1702301115163807980?s=20).
```
Authorize.net testkeskkonna kaardimakse vastu võetud: Mastercard
```

It's a difficult problem to solve with gettext, though - if not impossible, because gettext does not support much beyond pluralization rules. The most robust solution would be to avoid using placeholders, and always write out full sentences (not something I'm actually recommending to do at the moment though).

Another option is to use something like Fluent or MessageFormat, which have much better support for inflections, cases, etc - but they are foreign to the WP ecosystem, unfortunately.

itambek-godaddy (approved):
LGTM, i18n is hard 😅

woocommerce/Lifecycle.php
unfulvio-godaddy: they had a question about "Hot dog"! I LOL'ed -- taken out of context is funny

woocommerce/payment-gateway/class-sv-wc-payment-gateway-direct.php
unfulvio-godaddy: in the end this is a quasi-machine message, more of a notification for the merchant running a test so I think we can accept some degree of linguistic inaccuracy

PR #641, opened 2023-11-10
Release: v5.11.11

unfulvio-godaddy:
## Summary

This release cherry picks some changes from #614 to update how supported features from WC are handled and to add a basic `Blocks_Handler` to ensure we can declare the plugin's compatibility with blocks and display a notice otherwise. 

Plugins implementing this version of the FW need to mainly change the `supports_hpos` argument in the main plugin file constructor to a `supported_features` param, which by default will be assumed as follows:

```php
 'supported_features' => [
	'hpos'   => false,
	'blocks' => [
		'cart'     => false,
		'checkout' => false,
	],
],
```

If a plugin is compatible with HPOS, they need to change the corresponding key to `true` value. Same with either the cart/checkout blocks.

If a plugin is not compatible with either block, dismissble notices like the following will be produced _if_ either block is found to be active in the respective cart/checkout pages:

image

No other action is required in the plugin implementing this version other than QA.

## QA

The following QA applies to all plugins, so I'm placing the steps here as a catch-all, with notes added on a per-plugin basis when necessary.

#### Setup

- Use WooCommerce version 8.3-RC1 or higher 
- Switch to the plugin branch utilizing this FW version

#### Steps

1. You have an install using the cart/checkout shortcodes
    - [x] The notice won't appear, plugin functionality unchanged
1. You have an install using the cart/checkout blocks instead
    - [x] The notice(s) will appear
1. Click on the documentation links of either notice 
    - [x] You are redirected in the appropriate doc page in a new browser window
1. Click on the CTA button
    - [x] You are redirected to the corresponding edit screen in the same window
    - [x] A WooCommerce tool to help you revert the block to shortcode should be present
1. Test the notice behavior by reverting either incompatible block to shortcode manually
    - [x] The corresponding notice disappears
1. Dismiss a notice
    - [x] The notice is dismissed permanently regardless of block present or not

### Before merge

- [x] Please also check https://github.com/godaddy-wordpress/wc-plugin-framework/pull/642 as it's a simple change that could be rolled out into this PR


woocommerce/class-sv-wc-plugin.php
```diff
@@ -106,13 +111,13 @@ abstract class SV_WC_Plugin {
...
-	 * @param array $args {
+	 * @param array<string, mixed> $args {
 	 *     optional plugin arguments
 	 *
 	 *     @type int|float $latest_wc_versions the last supported versions of WooCommerce, as a major.minor float relative to the latest available version
 	 *     @type string $text_domain the plugin textdomain, used to set up translations
-	 *     @type bool $supports_hpos whether the plugin supports HPOS (default false)
-	 *     @type array  $dependencies {
+	 *     @type array<string, mixed> $supported_features supported WooCommerce features (HPOS, Blocks, etc.)
+	 *     @type array<string, array<string, mixed>>  $dependencies {
```
itambek-godaddy, line 120: What do you think about using an array shape do describe the arguments, so that we'd have IDE autocompletion support as well as PHPStan static type checking support?

Whenever I need to update the plugin constructor arguments in individual plugins, I find that I need to go back to the abstract constructor to figure out what it accepts. And even so, it's not clear which "features" I can declare support for.

```php
/**
 * @param array{
 *     latest_wc_versions?: int|float,
 *     text_domain?: string,
 *     supported_features?: array{
 *         hpos?: bool,
 *         blocks?: array{
 *             cart?: bool,
 *             checkout?: bool
 *         }
 *     },
 *     dependencies?: array{
 *         php_extensions?: array<string, mixed>,
 *         php_functions?: array<string, mixed>,
 *         php_settings?: array<string, mixed>
 *     }
 * } $args
 */
```

We could define the various parts as local type aliases`@phpstan-type` in the class doblock and then reuse them, when needed, for example:

```php
/**
 * @phpstan-type SupportedFeatures array{ hpos?: bool, blocks?: array{ cart?: bool, checkout?: bool }}
 */
abstract class SV_WC_Plugin {

	/** @var SupportedFeatures */
	protected $supported_features = [];

	/**
	 * @return SupportedFeatures
	 */
	public function get_supported_features() : array {
		return $this->supported_features;
	}
}
```

woocommerce/Blocks/Blocks_Handler.php
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		return CartCheckoutUtils::is_cart_block_default();
+	}
+
+
+	/**
+	 * Adds admin notices pertaining the blocks integration.
+	 *
+	 * @since 5.12.0
```
itambek-godaddy, line 111:
```suggestion
	 * @since 5.11.11
```

itambek-godaddy (changes requested):
Looks good, there are a few incorrect version tags, which need to be addressed and I also have a general suggestion about the typing the array shapes for features and other constructor arguments.

itambek-godaddy (approved):
LGTM, we can revisit the types in 5.12

PR #651, opened 2023-12-07
Allow plugins to hook into the transaction response

davidperezgar:
We have an action that need the transaction response and send to an API software that process the transaction. Actually is hardcoded and it's not the best way to do it. Better to make as WordPress way, with do_action hook. This line is good for other developers that wants to hook in that line of code.

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-hosted.php
@@ -406,6 +406,9 @@ abstract class SV_WC_Payment_Gateway_Hosted extends SV_WC_Payment_Gateway {
 			// Handle the order based on the response
 			$this->process_transaction_response( $order, $response );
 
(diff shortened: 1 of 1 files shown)


unfulvio-godaddy:
hey @davidperezgar thanks for contributing to the framework and sorry for the late reply!

I am going to close this but I have added your suggested hook in the upcoming 5.12.0 version of the framework due to be released later today: https://github.com/godaddy-wordpress/wc-plugin-framework/pull/614/commits/6164f80ece08d86c96693854627f01d83bdaef54 - I have changed the hook name slightly to be dynamically built based on the gateway name, consistently with how we do for the rest of these hooks. Hope this will be helpful to your and your team. Please feel free to reopen this issue or open another issue if you need to report something or have a similar request.

davidperezgar:
Hey! That's awesome! Thanks! I'm using the plugin Global Payments HPP. What would be that ID?

PR #657, opened 2024-01-25
Add methods to work with the payment processing context

itambek-godaddy:
# Summary

In the 5.12.0 release, we added "context" to the payment data, both in shortcode and block checkout.
However, we did not add any methods that plugins could use to determine the context easily.

This PR adds a few helper methods to the gateway class to determine the current processing context.

## Details

We did have a `SV_WC_Payment_Gateway::is_block_checkout()` method, which used a POST value that we were using while developing 5.12.0. However, at some point we changed the POST data, and forgot to update to work with the new context field.

I renamed the existing `SV_WC_Payment_Gateway::is_block_checkout()` method, because it's only used in one plugin, which hasn't been released yet.

I used the term "processing context" because "payment context" could easily be confused with the `PaymentContext` passed to `Gateway_Checkout_Block_Integration::prepare_payment_data()`. Also, "checkout context" did not seem precise enough, because "checkout" can either refer to the checkout page or the action of checking out.

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
@@ -296,7 +296,9 @@ class SV_WC_Payment_Gateway_Payment_Form extends Handlers\Script_Handler {
 	 */
 	protected function get_payment_fields() : array {
 
-		switch ( $this->get_gateway()->get_payment_type() ) {
+		$gateway = $this->get_gateway();
+
+		switch ( $gateway->get_payment_type() ) {
 
 			case 'credit-card':
 				$fields = $this->get_credit_card_fields();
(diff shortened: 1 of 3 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -4382,14 +4386,41 @@ public function is_echeck_gateway() {
...
+
+	/**
+	 * Checks if we're currently in teh given payment processing context.
+	 *
+	 * @since 5.12.1-dev.1
 	 * @see Gateway_Checkout_Block_Integration::prepare_payment_data()
 	 *
 	 * @return bool
 	 */
-	protected function is_block_checkout(): bool {
```
unfulvio-godaddy, line 4390: just concerned if we are accessing this method from any of the gateways we've been releasing so far

unfulvio-godaddy (approved):

woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
itambek-godaddy: Cybersource should be the only one, and it hasn't been released.

woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -339,7 +341,7 @@ protected function get_payment_fields() : array {
 		 * @param array<string, mixed> $fields in the format supported by {@see woocommerce_form_field()}
 		 * @param SV_WC_Payment_Gateway_Payment_Form $payment_form payment form instance
 		 */
-		return (array) apply_filters( 'wc_' . $this->get_gateway()->get_id() . '_payment_form_default_payment_form_fields', $fields, $this );
```
unfulvio-godaddy, line 342: maybe we should continue calling `get_gateway()` here?

woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
unfulvio-godaddy: alright then, latest commit looks good too

PR #659, opened 2024-02-05
Add support for dynamic properties to improve compatibility with newer PHP versions

unfulvio-godaddy:
Adds a `#[\AllowDynamicProperties]` declaration at the beginning of all classes in the framework. This is a bit of a blanket solution to address certain deprecation notices related to PHP 8.2 but that may be triggered in frameworked plugins. Note that core WP adopted a similar strategy. Individual plugins may still need to address compatibility if they use dynamic properties in their own classes.

## QA

- [x] Code review


itambek-godaddy (approved):
LGTM, thanks!

PR #660, opened 2024-02-15
 Add the checkout button text to the payment data payload 

unfulvio-godaddy:
## Summary

Adds the checkout button text to the payment data payload that may be used by the FE scripts. This in turn will make sure we hit a filter to let merchants customize that button text.

### Story [MWC-15558](https://godaddy-corp.atlassian.net/browse/MWC-15558)


sporter1-godaddy (approved):
Looks great to me! I was able to test this out via blocks, and can see and use the value coming through perfectly:
CleanShot 2024-02-16 at 11 45 52@2x

PR #662, opened 2024-02-19
Fix `remove_support()` method in base gateway class

nikolas4175-godaddy:
# Summary

Prevents `SV_WC_Payment_Gateway::remove_support()` from erroneously removing the item at index 0 when `$name` is not found (i.e. feature is already not supported) and also from creating gaps in gateway `supports` array.

[Slack discovery](https://godaddy.slack.com/archives/C06748WQS2Z/p1708347025232569?thread_ts=1708092348.195799&cid=C06748WQS2Z) for context (s/o @sporter1-godaddy).

### Issue: [MWC-15631](https://godaddy-corp.atlassian.net/browse/MWC-15631?atlOrigin=eyJpIjoiMzkxMmFkNmYxZWJiNDVhMTk3MGMxOTI4MmFkMWZmZmYiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

# QA
- [ ] Attempting to remove an already-unsupported feature does not change the `supports` array
- [ ] Removing supported features works as expected
- [ ] When a feature is removed, the `supports` array is not left with any index gaps

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -4083,20 +4083,25 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 
 		foreach ( $feature as $name ) {
 
-			unset( $this->supports[ array_search( $name, $this->supports ) ] );
+			// Bail if the feature already isn't supported
+			if( $key = array_search( $name, $this->supports ) !== false ){
 
-			/**
-			 * Payment Gateway Remove Support Action.
-			 *
-			 * Fired when removing support for a specific gateway feature. Allows other actors
-			 * (including ourselves) to take action when support is removed.
-			 *
-			 * @since 4.1.0
-			 *
-			 * @param SV_WC_Payment_Gateway $this instance
-			 * @param string $name of supported feature being removed
-			 */
-			do_action( 'wc_payment_gateway_' . $this->get_id() . '_removed_support_' . str_replace( '-', '_', $name ), $this, $name );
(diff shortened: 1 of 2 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -4081,22 +4081,34 @@ public function remove_support( $feature ) {
 			$feature = array( $feature );
 		}
 
+		$did_remove = false; 
+
 		foreach ( $feature as $name ) {
 
-			unset( $this->supports[ array_search( $name, $this->supports ) ] );
+			// Bail if the feature already isn't supported
+			if( $key = array_search( $name, $this->supports ) !== false ){
```
unfulvio-godaddy, line 4089: this might cause `$key`  to be boolean

woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -4081,22 +4081,34 @@ public function remove_support( $feature ) {
...
+				 * @param string $name of supported feature being removed
+				 */
+				do_action( 'wc_payment_gateway_' . $this->get_id() . '_removed_support_' . str_replace( '-', '_', $name ), $this, $name );
+			}
+		}
+
+		if( $did_remove ){
+			// re-index to clean up any new gaps
+			$this->supports = array_values( $this->supports );
 		}
```
unfulvio-godaddy, line 4112: no need this condition, we can always re-index

note the spacing with parenthesis is off (both in WP, SkyVerge or PSR they require a space)

unfulvio-godaddy (approved):

PR #666, opened 2024-03-12
Add non-gateway cart and checkout blocks supporting logic 🔥 😈 🔥 

unfulvio-godaddy:
## Summary

Adds non-gateway cart and checkout blocks supporting logic.

### Base PR: #665


itambek-godaddy (approved):
LGTM

PR #667, opened 2024-03-15
Update npm build packages

### LINKED ISSUE #360: Source map is not included in the package but is referenced by the min file
See https://wordpress.org/support/topic/file-missing-from-plugin-2/ - Safari logs the missing source map as an error to the dev console.

The map files should either be included in this package, or the JS should be compiled without source maps.

unfulvio-godaddy:
## Summary

This PR updates npm dependencies that are used in the FW to build its static assets, generate the potfile, compile scss and coffeescript, webpack, etc.

This should address some security warnings although the dependencies per se are not part of the distribution so there are no security risks involved.

I was hoping it would have addressed the issue with #360 however it turns out the output is still problematic and likely an issue with the `grunt-contrib-sass` package. I'll open an issue there to ask for configuration options (update: issue opened https://github.com/gruntjs/grunt-contrib-sass/issues/283). 

For now I have manually edited the affected map file references, but there changes would be reverted if we are to build again those css files (does not happen that frequently).

### Closes #360 (see above caveats though, will require follow up)

would also close several open prs:  https://github.com/godaddy-wordpress/wc-plugin-framework/pulls?q=is:pr+is:open+sort:updated-desc

## QA

To test these changes, the reviewer should run this version of the FW with a gateway that uses the payment form and account payment methods, ensuring the functionality and CSS is unaltered.


itambek-godaddy (approved):
I used this branch on Cybersource, and was able to checkout and add payment methods on the My Account page successfully. No visual differences detected, either.

PR #671, opened 2024-03-21
Update class-sv-wp-background-job-handler.php

ramonfincken:
Make this more efficient. 

While awaiting a dedicated database table where no LIKE queries will be needed or at least be efficient in a small colum instead of full search in option_value as now.

https://github.com/godaddy-wordpress/wc-plugin-framework/issues/672


unfulvio-godaddy (approved):
this looks like a good addition, thank you!

for now we are relaying on the wp_options table as you've noticed but the points you've raised in #672 are legit

PR #676, opened 2024-05-01
Use gateway FW path when loading block handler files

itambek-godaddy:
# Summary

Some FW files only exist in gateway plugins due to our build process. This change ensures that gateway plugins will load block handler files from the gateway FW path, rather than a generic FW path.

This fix is similar to https://github.com/gdcorp-partners/woocommerce-gateway-intuit-qbms/pull/11/files

### Story: [MWC-16371](https://godaddy-corp.atlassian.net/browse/MWC-16371)

### DIFF
--- a/woocommerce/payment-gateway/Blocks/Gateway_Blocks_Handler.php
+++ b/woocommerce/payment-gateway/Blocks/Gateway_Blocks_Handler.php
@@ -59,7 +59,7 @@ class Gateway_Blocks_Handler extends Blocks_Handler {
 			/** @var SV_WC_Payment_Gateway_Plugin $plugin */
 			$plugin = $this->plugin;
 
-			require_once( $plugin->get_payment_gateway_framework_path() . '/payment-gateway/Blocks/Gateway_Checkout_Block_Integration.php' );
+			require_once( $plugin->get_payment_gateway_framework_path() . '/Blocks/Gateway_Checkout_Block_Integration.php' );
 
 			foreach ( $plugin->get_gateways() as $gateway ) {
 
--- a/woocommerce/payment-gateway/Blocks/Gateway_Blocks_Handler.php
+++ b/woocommerce/payment-gateway/Blocks/Gateway_Blocks_Handler.php
@@ -59,7 +59,7 @@ class Gateway_Blocks_Handler extends Blocks_Handler {
 			/** @var SV_WC_Payment_Gateway_Plugin $plugin */
 			$plugin = $this->plugin;
 
-			require_once( $plugin->get_framework_path() . '/payment-gateway/Blocks/Gateway_Checkout_Block_Integration.php' );
+			require_once( $plugin->get_payment_gateway_framework_path() . '/payment-gateway/Blocks/Gateway_Checkout_Block_Integration.php' );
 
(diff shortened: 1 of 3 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway-plugin.php
```diff
@@ -181,8 +181,8 @@ protected function init_rest_api_handler() {
 	 */
 	protected function init_blocks_handler() : void {
 
-		require_once( $this->get_framework_path() . '/Blocks/Blocks_Handler.php' );
-		require_once( $this->get_framework_path() . '/payment-gateway/Blocks/Gateway_Blocks_Handler.php' );
+		require_once( $this->get_payment_gateway_framework_path() . '/Blocks/Blocks_Handler.php' );
```
nikolas4175-godaddy, line 184:
```suggestion
		require_once( $this->get_framework_path() . '/Blocks/Blocks_Handler.php' );
```
AFAICT the base `Blocks_Handler` file doesn't exist under `/payment-gateway/` so this one should still load from the FW root dir.

woocommerce/payment-gateway/Blocks/Gateway_Blocks_Handler.php
```diff
@@ -59,7 +59,7 @@ public function handle_blocks_integration() : void {
 			/** @var SV_WC_Payment_Gateway_Plugin $plugin */
 			$plugin = $this->plugin;
 
-			require_once( $plugin->get_framework_path() . '/payment-gateway/Blocks/Gateway_Checkout_Block_Integration.php' );
+			require_once( $plugin->get_payment_gateway_framework_path() . '/payment-gateway/Blocks/Gateway_Checkout_Block_Integration.php' );
```
nikolas4175-godaddy, line 62:
```suggestion
			require_once( $plugin->get_payment_gateway_framework_path() . '/Blocks/Gateway_Checkout_Block_Integration.php' );
```
`/payment-gateway` is included in the `get_payment_gateway_framework_path()` return value

woocommerce/payment-gateway/class-sv-wc-payment-gateway-plugin.php
```diff
@@ -181,8 +181,8 @@ protected function init_rest_api_handler() {
...
+		require_once( $this->get_payment_gateway_framework_path() . '/payment-gateway/Blocks/Gateway_Blocks_Handler.php' );
```
nikolas4175-godaddy, line 185:
```suggestion
		require_once( $this->get_payment_gateway_framework_path() . '/Blocks/Gateway_Blocks_Handler.php' );
```
Same as above, no need to repeat `/payment-gateway`

nikolas4175-godaddy (changes requested):

itambek-godaddy:
@nikolas4175-godaddy thanks for the review and the catch in the path. I did test this with auth.net to see if I would get any fatal errors, but I think I had some other gateway plugin active as well, so the code _seemed_ to work. My apologies, shouod have been more thorough.

nikolas4175-godaddy (approved):
Updates LGTM!

PR #677, opened 2024-05-30
Release 5.12.4

itambek-godaddy:
# Summary

This release addresses an issue discovered in the 5.12.3 release, where bundled gateway scripts trigger errors.

## Details

This was a very enigmatic issue to track down. Recently, our [NPM deps were updated](https://github.com/godaddy-wordpress/wc-plugin-framework/commit/0e25ac63f18501a502a7a2a4b0ba215aef5e45b7), which included a major upgrade to `parcel` - from v1 to v2. Parcel v2 compiles code to JS modules and requires the `type="module"` attribute on the script tags when loading. Because the compiled assets were loaded as regular JS, it caused a `TypeError` when the script was executed.

Unfortunately, WP only added `wp_enqueue_script_module` in 6.5 (this year), and adding the `type` attribute in earlier WP versions requires [some filtering](https://stackoverflow.com/a/77863823). Even when I tried adding the type attribute, I ran into other issues.

The change to using modules requires far more QA than downgrading parcel at the moment, which is why this PR has chosen the latter. I suggest we circle back to upgrading parcel later when we have more time to dedicate to it.

## QA

- 1. In one of the gateway plugins (such as ANet), require this branch of the FW
- 2. Go to My Account > Payment Methods
- 3. Click on "Edit" on any of the payment methods
- [ ] The edit form works, edits are saved and there are no errors in console.


package.json
```diff
@@ -34,7 +33,7 @@
     "grunt-shell": "^4.0",
     "grunt-text-replace": "^0.4",
     "load-grunt-tasks": "~5.1",
-    "parcel": "^2.12",
+    "parcel": "1.12.3",
```
nikolas4175-godaddy, line 36: @itambek-godaddy it looks like there's a v1.12.5 [available on npm](https://www.npmjs.com/package/parcel-bundler) under the `parcel-bundler` name. Should we be using that instead?

nikolas4175-godaddy (approved):
## ✅  QA
- In ANet, require this branch of the FW
- Go to My Account > Payment Methods
- Click on "Edit" on any of the payment methods
- [x] The edit form works, edits are saved and there are no errors in console.
---
Left a quick note about the specific Parcel version we're rolling back to, but approving anyhow as the current changes resolve the known build issue.

package.json
itambek-godaddy: Ah, right - it's because starting `1.12.4` there is [an issue with the build](https://github.com/parcel-bundler/parcel/issues/5943). The [workaround mentioned here](https://github.com/parcel-bundler/parcel/issues/5943#issuecomment-788928442) did not work, so I opted for downgrading instead.

PR #685, opened 2024-06-14
Disable order saving when we're on the payment method page

agibson-godaddy:
# Summary

This prevents us from saving the order object when we're on the "add payment method" page. We do this by ensuring we don't call any save methods if the order doesn't already have an ID number. When on the payment method page, we expect to be working with a `WC_Order` object that has a `0` ID which is why this works.

## Details

The initial `$order->save()` method has been removed. This was added to avoid potential errors, but thus far we haven't been able to reproduce any errors with that line removed.

Using authorize.net I tested the following:

- Placing an order, using an existing/saved payment method; and
- Placing an order, using a brand new payment method

In both cases the order was created successfully and the charge went through.

### Authorize.net

See https://github.com/gdcorp-partners/woocommerce-gateway-authorize-net-cim/pull/98

1. Ensure you have authorize.net set up.
2. Add a product to your cart and proceed through to checkout.
3. Select the authorize.net payment gateway and add a new payment method. Complete the checkout.
    - [ ] A new order is created with all appropriate information.
    - [ ] Payment is taken successfully.
5. Go to your account area on the front-end and to the "payment methods" page. Click through to add a new payment method and complete the process.
6. Once added, log in as an admin and check the WooCommerce > Orders.
    - [ ] No new order has been created.

### Elavon

See https://github.com/gdcorp-partners/woocommerce-gateway-elavon/pull/94

1. Ensure you have Elavon set up and enabled.
2. Visit any page on the front end.
    - [ ] No new order is created.
3. Add a product to your cart and proceed through to checkout.
4. Select the Elavon payment gateway and complete your payment.
    - [ ] One new order is created with all appropriate information.
    - [ ] Payment is taken successfully.
5. Go to your account area on the front-end and to the "payment methods" page. Click through to add a new payment method and complete the process.
6. Once added, log in as an admin and check the WooCommerce > Orders.
    - [ ] No new order has been created.

## Before Merge

- [ ] Framework namespace and version is updated

### DIFF
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -1949,6 +1949,7 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 			$order->save();
 		}
 
+		// the get_order_with_unique_transaction_ref() call results in saving the order object, which we don't want to do on the "add payment method" page
 		if ( ! is_add_payment_method_page() ) {
 			$order = $this->get_order_with_unique_transaction_ref($order);
 		}
--- a/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
+++ b/woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
@@ -1944,8 +1944,8 @@ abstract class SV_WC_Payment_Gateway extends \WC_Payment_Gateway {
 		/* translators: Placeholders: %1$s - site title, %2$s - order number */
 		$order->description = sprintf( esc_html__( '%1$s - Order %2$s', 'woocommerce-plugin-framework' ), wp_specialchars_decode( SV_WC_Helper::get_site_name(), ENT_QUOTES ), $order->get_order_number() );
 
-		// the get_order_with_unique_transaction_ref() call results in saving the order object, which we don't want to do on the "add payment method" page
-		if ( ! is_add_payment_method_page() ) {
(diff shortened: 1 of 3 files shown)


woocommerce/payment-gateway/class-sv-wc-payment-gateway.php
```diff
@@ -1944,13 +1944,11 @@ public function get_order( $order ) {
 		/* translators: Placeholders: %1$s - site title, %2$s - order number */
 		$order->description = sprintf( esc_html__( '%1$s - Order %2$s', 'woocommerce-plugin-framework' ), wp_specialchars_decode( SV_WC_Helper::get_site_name(), ENT_QUOTES ), $order->get_order_number() );
 
-		// when HPOS is enabled, we need to save the order to avoid potential errors when saving a new payment method
-		if ( SV_WC_Plugin_Compatibility::is_hpos_enabled() && ( ! is_admin() || is_ajax() ) ) {
-			$order->save();
+		// the get_order_with_unique_transaction_ref() call results in saving the order object, which we don't want to do if the order hasn't already been saved (such as when adding a payment method)
+		if ( $order->get_id() ) {
```
agibson-godaddy, line 1948: The previous fix of checking if we were on the "Add Payment Method" page did not work. The problem was that it would fail if you were on the payment method page, but doing an ajax action (like Elavon does). So it would create orders during the ajax processing.

agibson-godaddy:
QA done here: https://godaddy.slack.com/archives/C04L7S7JJE9/p1718716914136279?thread_ts=1718233890.103419&cid=C04L7S7JJE9

PR #690, opened 2024-07-19
Authorize.Net enqueues jQuery in the footer in fall back scenarios [1]

nmolham-godaddy:
# Summary

Make sure payment gateway form script dependencies are rendered/loaded first.

## Details

- `wp_print_scripts` doesn't re-render/re-print already loaded scripts, it does an internal check.
- `wp_print_scripts` flags done script, so it won't re-do them again.
- Added `SV_WC_Payment_Gateway_Payment_Form::getFormScriptDependencies` method to allow gateways to define their own dependencies by overriding the method. By default all gateways require jQuery at least.

## QA

1. Add the following code snippet right before the `switch` block in `SV_WC_Payment_Gateway_Payment_Form::render_js` method
    - `wp_dequeue_script('jquery');`
2. Update the Authorize.Net plugin `woocommerce-gateway-authorize-net-cim` composer.json to use this branch of the framework.
3. Load any frontend page
    - [ ] JQuery script tag is render right before the form main script.
    - [ ] No error `jQuery is not defined` logged in the console.

- [x] Code review

## Before merge


nmolham-godaddy:
@agibson-godaddy, I wasn't able to reproduce it either, but I simulated it manually as the 2nd best thing. 

> Worst case scenario, I don't mind just approving this anyway because it does make sense and I confirmed AuthorizeNet still works with it.

I think so too, unless there could be a possible side effect that I am not aware of! CC @astandiford-godaddy @ajaynes-godaddy @nikolas4175-godaddy @itambek-godaddy

agibson-godaddy (approved):
Thinking we should just merge this. I don't think there are any pitfallls.

PR #696, opened 2024-08-16
Set the payment form context in the session

agibson-godaddy:
# Summary

Related: https://github.com/gdcorp-partners/woocommerce-gateway-elavon/pull/103

This sets the page context in the session when rendering the payment forms.

## Details

- Also copied the `EnumTrait` from `mwc-common`
- Set up an enum for the contexts

## QA

_If applicable, add specific steps for the reviewer to perform as part of their QA process prior to approving this pull request. Steps should be in a step -> success? format, like below_

### Setup

- [x] QA in https://github.com/gdcorp-partners/woocommerce-gateway-elavon/pull/103 passes

## Before merge


woocommerce/payment-gateway/class-sv-wc-payment-gateway-payment-form.php
```diff
@@ -991,6 +992,31 @@ public function render_payment_fields() {
 		foreach ( $this->get_payment_fields() as $field ) {
 			$this->render_payment_field( $field );
 		}
+
+		// set the context for the checkout form in case gateways need to reference this in their validation
+		if ($context = $this->getCurrentPaymentFormContext()) {
+			WC()->session->set(
+				sprintf('wc_%s_payment_form_context', $this->get_gateway()->get_id()),
```
nmolham-godaddy, line 999: We can use the extracted `paymentFormContextKeyName` here, ref: https://github.com/gdcorp-partners/woocommerce-gateway-elavon/pull/103#discussion_r1721181978

agibson-godaddy: See updates here: https://github.com/godaddy-wordpress/wc-plugin-framework/pull/696/commits/34b27ababe52c14eb16f48d4e90f92628ca6f490

I created a new class for all the logic because I realized that the payment form class isn't accessible in Elavon where we actually want to use this (that's in the `AJAX` class).

nmolham-godaddy: Good idea, nice

nmolham-godaddy:
@agibson-godaddy So, I got carried away a bit here 😅 

At first, I wanted to add unit test for the new `PaymentFormContextChecker` component, so I ended up adding `mwc-tests`, setup PHPUnit and converting existing codeception unit test to PHPUnit tests along side adding the new class tests 

Ref: 34b27ab...dc8f67c

I am not sure why Github doesn't see the `6792-mwcmanaged` runner and the job fails straight away instead of the usual hold/pending status. Maybe because this repo is on a different organization level `godaddy-wordpress` instead of `gdcorp-partners` ? 🤔

agibson-godaddy:
> @agibson-godaddy So, I got carried away a bit here 😅
> 
> At first, I wanted to add unit test for the new `PaymentFormContextChecker` component, so I ended up adding `mwc-tests`, setup PHPUnit and converting existing codeception unit test to PHPUnit tests along side adding the new class tests
> 
> Ref: [34b27ab...dc8f67c](https://github.com/godaddy-wordpress/wc-plugin-framework/compare/34b27ababe52c14eb16f48d4e90f92628ca6f490...dc8f67c2b93c9327f453ff01aa90f8c2ab676f2f)
> 
> I am not sure why Github doesn't see the `6792-mwcmanaged` runner and the job fails straight away instead of the usual hold/pending status until the repo gets added to the pool's repo list. Maybe because this repo is on a different organization level `godaddy-wordpress` instead of `gdcorp-partners` ? 🤔

That could definitely be the problem. Runner pools are per-organization, and every repo added is assumed to be part of that org. So we can't even add this repo to our existing runner pool because of that. We'd have to create a separate runner pool.

I don't know for certain that's what's causing the instant error, but it's definitely a problem anyway that we have to fix!

I can take a look at this today to see if I can get it running. Nice work on the tests!!! 🙌

agibson-godaddy:
@nmolham-godaddy Reworked to drop `mwc-tests` dependency and use hosted runners.

nmolham-godaddy (approved):
LVGTM

PR #703, opened 2024-08-23
[WIP] Add `CanConvertToArrayTrait` and `ArrayHelper` classes

agibson-godaddy:
# Summary

Release: https://github.com/godaddy-wordpress/wc-plugin-framework/pull/699

The main goal of this PR was to add the `CanConvertToArrayTrait`. Our existing implementation of this uses several methods from `ArrayHelper` so I copied over those as well. For now I only copied over the methods the trait actually needs. I'm not opposed to adding the rest, I just did it like this because I was also rewriting the tests (ones in mwc-common are kind of old) so I was reducing the amount of work required there in one go.

Note that the framework does already have some array helpers [starting here](https://github.com/godaddy-wordpress/wc-plugin-framework/blob/master/woocommerce/class-sv-wc-helper.php#L270). I propose we eventually follow up with another PR to start moving those methods over to the new `ArrayHelper` class (forwarding the old methods to the new for backwards compatibility). Again just didn't do that in this PR to keep this one smaller.

## QA

_If applicable, add specific steps for the reviewer to perform as part of their QA process prior to approving this pull request. Steps should be in a step -> success? format, like below_

### Setup

_List any configuration requirements for testing_

- This setting is configured
- Taxes are enabled

### Steps

Steps should be in a step => success? format, like below:

1. Do this
1. And this
	- [ ] The result is this

- [ ] Code review

## Before merge

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -9,6 +9,8 @@ require_once PLUGIN_ROOT_DIR.'/woocommerce/class-sv-wc-plugin.php';
 require_once PLUGIN_ROOT_DIR.'/woocommerce/class-sv-wc-plugin-exception.php';
 require_once PLUGIN_ROOT_DIR.'/woocommerce/Enums/Traits/EnumTrait.php';
 require_once PLUGIN_ROOT_DIR.'/woocommerce/Enums/PaymentFormContext.php';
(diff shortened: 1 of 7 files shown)


nmolham-godaddy (approved):
Nice work, I finished off the rest of the TODO tests

agibson-godaddy:
Thank you!!! Nice work 🎉

PR #717, opened 2024-09-25
Account for Subscriptions Core library when checking for Subscriptions

unfulvio:
# Summary

WooCommerce Subscriptions might exist as a split library from https://github.com/Automattic/woocommerce-subscriptions-core -- this library includes nearly all the functionalities from WooCommerce Subscriptions. 

If a plugin or third party code uses this library, gateways or plugins that use this framework and integrate with Subscriptions won't work because this framework only checks if the Subscriptions plugin is active, not the standalone lib.

The checks in this PR have been updated to account for that possibility.

### DIFF
--- a/woocommerce/class-sv-wc-plugin-compatibility.php
+++ b/woocommerce/class-sv-wc-plugin-compatibility.php
@@ -324,22 +324,31 @@ class SV_WC_Plugin_Compatibility {
 	 */
 	protected static function get_wc_subscriptions_version() : ?string {
 
+		$version = null;
+
 		if ( class_exists( 'WC_Subscriptions' ) && ! empty( \WC_Subscriptions::$version ) ) {
-			return \WC_Subscriptions::$version;
+			$version = \WC_Subscriptions::$version;
 		}
 
-		if ( class_exists( 'WC_Subscriptions_Core_Plugin' ) ) {
+		if ( ! $version && class_exists( 'WC_Subscriptions_Core_Plugin' ) ) {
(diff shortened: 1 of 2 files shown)


woocommerce/class-sv-wc-plugin-compatibility.php
```diff
@@ -322,9 +322,33 @@ public static function is_wc_subscriptions_version_lt( $version ) {
...
+	protected static function get_wc_subscriptions_version() : ?string {
 
-		return class_exists( 'WC_Subscriptions' ) && ! empty( \WC_Subscriptions::$version ) ? \WC_Subscriptions::$version : null;
+		$version = null;
+
+		if ( class_exists( 'WC_Subscriptions' ) && ! empty( \WC_Subscriptions::$version ) ) {
+			$version = \WC_Subscriptions::$version;
+		}
+
+		if ( ! $version && class_exists( 'WC_Subscriptions_Core_Plugin' ) ) {
```
nmolham-godaddy, line 333:
```suggestion
		if ( class_exists( 'WC_Subscriptions' ) && ! empty( \WC_Subscriptions::$version ) ) {
			$version = \WC_Subscriptions::$version;
		} else if ( ! $version && class_exists( 'WC_Subscriptions_Core_Plugin' ) ) {
```

Do you think this should be `else if` instead of completely override the version by the core package?

unfulvio: sure, it would work either way

nmolham-godaddy (approved):
LVGTM

PR #718, opened 2024-09-26
Deprecate Enhanced navigation feature integration

nmolham-godaddy:
# Summary

Removes support for the deprecated Enhanced navigation feature.

## QA

- [ ] Code review

## Before merge

### DIFF
--- a/woocommerce/class-sv-wc-helper.php
+++ b/woocommerce/class-sv-wc-helper.php
@@ -1048,8 +1048,7 @@ class SV_WC_Helper {
 	public static function is_wc_navigation_enabled() : bool
 	{
 		if (SV_WC_Plugin_Compatibility::is_wc_version_gte('9.3')) {
-			error_log('The Enhanced navigation feature has been deprecated since WooCommerce 9.3 with no alternative. Navigation classes will be removed in WooCommerce 9.4');
-
+			self::enhancedNavigationDeprecationNotice();
 			return false;
 		}
 
@@ -1058,6 +1057,11 @@ class SV_WC_Helper {
 			is_callable([\Automattic\WooCommerce\Admin\Features\Navigation\Menu::class, 'add_plugin_category']) &&
 			is_callable([\Automattic\WooCommerce\Admin\Features\Features::class, 'is_enabled']) &&
 			\Automattic\WooCommerce\Admin\Features\Features::is_enabled('navigation');
+	}
+
+	protected static function enhancedNavigationDeprecationNotice() : void
+	{
+		error_log('The Enhanced navigation feature has been deprecated since WooCommerce 9.3 with no alternative. Navigation classes will be removed in WooCommerce 9.4');
 	}
 
 
--- a/woocommerce/class-sv-wc-helper.php
+++ b/woocommerce/class-sv-wc-helper.php
@@ -1030,9 +1030,9 @@ class SV_WC_Helper {
 	 *
 	 * @return bool
 	 */
-	public static function is_enhanced_admin_screen() {
-
-		return is_admin() && SV_WC_Plugin_Compatibility::is_enhanced_admin_available() && ( \Automattic\WooCommerce\Admin\Loader::is_admin_page() || \Automattic\WooCommerce\Admin\Loader::is_embed_page() );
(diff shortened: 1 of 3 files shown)


woocommerce/class-sv-wc-helper.php
```diff
@@ -1041,16 +1041,15 @@ public static function is_enhanced_admin_screen() {
...
-			is_callable( [ \Automattic\WooCommerce\Admin\Features\Navigation\Screen::class, 'register_post_type' ] ) &&
-			is_callable( [ \Automattic\WooCommerce\Admin\Features\Navigation\Menu::class, 'add_plugin_item' ] ) &&
-			is_callable( [ \Automattic\WooCommerce\Admin\Features\Navigation\Menu::class, 'add_plugin_category' ] ) &&
-			is_callable( [ \Automattic\WooCommerce\Admin\Features\Features::class, 'is_enabled' ] ) &&
-			\Automattic\WooCommerce\Admin\Features\Features::is_enabled( 'navigation' );
+	public static function is_wc_navigation_enabled() : bool
+	{
+		error_log('The Enhanced navigation feature has been deprecated since WooCommerce 9.3 with no alternative. Navigation classes will be removed in WooCommerce 9.4');
+
+		return false;
```
agibson-godaddy, line 1052: Do you think we need a Woo version check here? [ref](https://godaddy.slack.com/archives/C01NDQZULKF/p1727279715494199?thread_ts=1727204166.496609&cid=C01NDQZULKF)

As in:

If on Woo GTE 9.3, then return false like you're doing here.
Otherwise, use previous code.

nmolham-godaddy: I can't think of any benefit for keeping the integration in the first place as it is already deprecated and will be removed pretty soon. I am leaning towards removing toward always returning false

woocommerce/class-sv-wc-helper.php
```diff
@@ -1041,16 +1041,32 @@ public static function is_enhanced_admin_screen() {
...
+	{
+		if (SV_WC_Plugin_Compatibility::is_wc_version_gte('9.3')) {
+			self::enhancedNavigationDeprecationNotice();
+			return false;
+		}
+
+		return self::isEnhancedNavigationFeatureEnabled();
+	}
+
+	protected static function isEnhancedNavigationFeatureEnabled() : bool
```
ajaynes-godaddy, line 1058: Missing a docblock here for the new method.
```suggestion
	/**
	 * Determines whether Woo's Enhanced Eavigation feature is enabled.
	 *
	 * @since 5.15.1
	 *
	 * @return bool
	 */	 
	protected static function isEnhancedNavigationFeatureEnabled() : bool
```

woocommerce/class-sv-wc-helper.php
```diff
@@ -1041,16 +1041,32 @@ public static function is_enhanced_admin_screen() {
...
+	{
+		return is_callable([\Automattic\WooCommerce\Admin\Features\Navigation\Screen::class, 'register_post_type']) &&
+			is_callable([\Automattic\WooCommerce\Admin\Features\Navigation\Menu::class, 'add_plugin_item']) &&
+			is_callable([\Automattic\WooCommerce\Admin\Features\Navigation\Menu::class, 'add_plugin_category']) &&
+			is_callable([\Automattic\WooCommerce\Admin\Features\Features::class, 'is_enabled']) &&
+			\Automattic\WooCommerce\Admin\Features\Features::is_enabled('navigation');
+	}
+
+	protected static function enhancedNavigationDeprecationNotice() : void
```
ajaynes-godaddy, line 1067: Another docblock
```suggestion
	/**
	 * Logs a notice for the Enhanced Navigation feature being deprecated.
	 *
	 * @since 5.15.1
	 *
	 * @return void
	 */
	protected static function enhancedNavigationDeprecationNotice() : void
```

ajaynes-godaddy (changes requested):
Couple of missing docblocks on new methods here. I couldn't remember if we can use a version placeholder that the build script replaces or not so I just went with the next patch release version for the `@since` annotations.

ajaynes-godaddy (approved):

PR #722, opened 2024-10-10
Cleaning up codebase (WIP)

nmolham-godaddy:
# Summary

Cleaning up and optimize the codebase.

## Details

There are so many aspects the codebase can be cleaned, so I thought about opening a PR to get feedback and inputs.

## QA

- [ ] Code review
- [ ] Checks pass

## Before merge


nmolham-godaddy:
@agibson-godaddy Agreed, these changes are are considering to be breaking changes. I was also thinking about raising the PHP min version to PHP 8.0, but I realised that would be unrealistic considering WooCommerce requirements.

PR #723, opened 2024-10-10
Add helper method to get WooCommerce object meta value

nmolham-godaddy:
# Summary

Adds `SV_WC_Helper::getWooCommerceObjectMetaValue` that gets value of a meta key from WooCommerce object based on its data type.

## QA

- [x] Code review
- [x] Checks pass

## Before merge


woocommerce/class-sv-wc-helper.php
```diff
@@ -1197,6 +1200,34 @@ public static function get_escaped_id_list( array $ids ) {
...
+	 *
+	 * @param WP_Post|WC_Data $object
+	 * @param string $field
+	 * @param bool $single
+	 *
+	 * @return array|mixed|string|null
+	 */
+	public static function getWooCommerceObjectMetaValue($object, string $field, bool $single = true)
+	{
+		$orderUtilExists = class_exists(OrderUtil::class);
```
agibson-godaddy, line 1214: Reading through the Woo source code, I understand what you have here will absolutely work for any WC class that extends `WC_Data` (such as products).

Just wondering if it's at all weird/risky to use the `OrderUtil` class anyway? That class is described as a utility class for orders, and it does work with products anyway, but do you think there's a risk of that ever changing so that it is orders only?

nmolham-godaddy: Good question, as far as I can tell, the same `OrderUtil` class is used in many places in WC codebase to get all sort of meta data for objects despite being orders or not! I know it is weird and sounds risky, but if you look into `OrderUtil::get_post_or_object_meta` source code, it doesn't case if the passed object is an order or not, it just loos for the `WC_Data` or `WP_Post` 

I can't find an equivalent to is within WC codebase for products or any other WC_Data/WP_Post based objects

ajaynes-godaddy: Knowing it's considered to be an internal utility class, it might not be the worst idea to abstract the source of `get_post_or_object_meta()` to our own trait or helper.

```php
	public function get_post_or_object_meta( ?WP_Post $post, ?\WC_Data $data, string $key, bool $single ) {
		if ( isset( $data ) ) {
			if ( method_exists( $data, "get$key" ) ) {
				return $data->{"get$key"}();
			}
			return $data->get_meta( $key, $single );
		} else {
			return isset( $post->ID ) ? get_post_meta( $post->ID, $key, $single ) : false;
		}
	}
```

nmolham-godaddy: At first I thought about doing so, but I remembered how many time WC change stuff internally that causes compatibility issues. So, the idea here is to use their implementation with fallback with our own implementation.

Make sense?

ajaynes-godaddy: I think I'm going to try just adding it as a second method to the helper, see how annoying the tests are to write.

ajaynes-godaddy: Added the compat method in 9f45812e46dbe561a6f2db73fec9e5f45b3a1036

Feel free to revert or adjust as needed.

nmolham-godaddy: That looks good to me, how about you @agibson-godaddy, what do you think?

agibson-godaddy (approved):
LGTM!

PR #736, opened 2025-02-17
Add a `PageHelper` class to check for current admin page

agibson-godaddy:
# Summary

This adds a `PageHelper` class with a handy method to check if we're on the WooCommerce analytics page.

## QA

- [x] QA steps in https://github.com/gdcorp-partners/woocommerce-cost-of-goods/pull/102 pass

## Before merge

### DIFF
--- a/tests/unit/Helpers/ArrayHelperTest.php
+++ b/tests/unit/Helpers/ArrayHelperTest.php
@@ -251,6 +251,92 @@ final class ArrayHelperTest extends TestCase
 		];
 	}
 
+	/**
+	 * Tests that the helper will return default with key not found.
+	 *
+	 * @covers ::get()
+	 */
+	public function testCanReturnDefaultWhenGetArrayValueByKeyNotFound() : void
+	{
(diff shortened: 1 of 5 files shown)


nmolham-godaddy (commented):
Nice addition, what do you think about this change [d0012c2](https://github.com/godaddy-wordpress/wc-plugin-framework/pull/736/commits/d0012c2fe56d58e5f3da216d7adb34cf8a7eab2f)? It makes sense to move the class sanity check into the `getWooCommercePageController` to avoid repeating that check in future methods

agibson-godaddy:
> Nice addition, what do you think about this change [d0012c2](https://github.com/godaddy-wordpress/wc-plugin-framework/pull/736/commits/d0012c2fe56d58e5f3da216d7adb34cf8a7eab2f)? It makes sense to move the class sanity check into the `getWooCommercePageController` to avoid repeating that check in future methods

You're right -- good call!

nmolham-godaddy (approved):
QA passed, I just took advantage of this opportunity to include ArrayHelper::get method as well, which made perfect sense in this context 7cff851...1c35f14

PR #745, opened 2025-03-04
Update workflow to remove deprecated usage of ubuntu-20.04 OS

agibson-godaddy:
# Summary

This switches to use a different version of Ubuntu for older PHP versions. As per https://github.com/actions/runner-images/issues/11101, Ubuntu 20.04 is being removed. In order to keep running tests against older PHP versions, this PR makes the following changes:

- Replaces Ubuntu 20.04 with a newer version
- The newer version doesn't include older versions of PHP, so we now have to use the setup PHP action in order to get older PHP versions available

## QA

- [ ] Code review
- [ ] Checks pass

## Before merge


ajaynes-godaddy (approved):
LGTM

PR #757, opened 2025-04-29
Fixes for loading translations too early

agibson-godaddy:
# Summary

Release: #755 

This is a partial fix (not guaranteed to be 100% complete) to address errors like this:

> Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the woocommerce-memberships domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. (This message was added in version 6.7.0.)

## Details

- `SV_WC_Hook_Deprecator` has been updated to accept the entire plugin instance as a parameter instead of just the plugin name. That way we can call `get_plugin_name()` when actually used instead of inside the constructor.

## QA

- [x] QA steps in https://github.com/gdcorp-partners/woocommerce-memberships/pull/1240 pass

## Before merge

### DIFF
--- a/woocommerce/class-sv-wc-hook-deprecator.php
+++ b/woocommerce/class-sv-wc-hook-deprecator.php
@@ -203,6 +203,12 @@ class SV_WC_Hook_Deprecator {
 		SV_WC_Helper::trigger_error( $message );
 	}
 
+
+	/**
+	 * Gets the plugin name.
+	 *
+	 * @return string
+	 */
 	protected function getPluginName() : string
 	{
 		if (isset($this->plugin)) {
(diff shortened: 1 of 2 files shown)


ajaynes-godaddy:
With this branch and the dev branch of the Framework, I'm still seeing the `_doing_it_wrong()` notices for memberships at least on the WooCommerce > Settings > Memberships page. Trying to track down what's triggering it.

ajaynes-godaddy:
Sorry, meant to leave that last comment on the memberships PR.

ajaynes-godaddy (approved):
LGTM

PR #760, opened 2025-04-30
Run 'clean' before build

agibson-godaddy:
# Summary

This update sthe workflow to run `npm run clean` before building. It seems occasionally when running `npm run build` the coffee assets aren't being compiled. This can be seen by looking at the commit column -- notice how some weren't changed!

Screenshot 2025-04-30 100543

After local testing I was able to reproduce those files not changing. They only finally changed by running `npm run clean` prior to building.

## QA

1. Run the "prep release" workflow from this branch. -- workflow run https://github.com/godaddy-wordpress/wc-plugin-framework/actions/runs/14751125116/job/41408663542
    - [x] Resulting PR contains all modified `dist/*` js files.

## Before merge

### DIFF
--- a/.github/workflows/prep-release.yml
+++ b/.github/workflows/prep-release.yml
@@ -61,6 +61,10 @@ jobs:
       - name: Bump package.json version
         run: npm version --no-git-tag-version --allow-same-version "${NEW_VERSION}"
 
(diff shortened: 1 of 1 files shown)


agibson-godaddy:
Here's a full list of modified files this time:

image

And here's the payment gateway form JS with the actual minified version:

image

